PR #4, opened 2021-04-03
Add ShellCheck

nanotee:
This PR adds ShellCheck to the list of available linters

The JSON object also has a `fix` key that contains the information necessary to do automatic fixes but AFAIK there's no way to configure client-side code actions.

Thanks for the cool plugin! Works really well :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ example on the `InsertLeave` or `TextChanged` events.
 
 - [Languagetool][5]
 - [Vale][8]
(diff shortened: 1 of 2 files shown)


mfussenegger:
Thank you for the PR.

> The JSON object also has a fix key that contains the information necessary to do automatic fixes but AFAIK there's no way to configure client-side code actions.

Yep that's right. I think it's also out-of-scope for this plugin to provide fixer/code-action capabilities.

PR #7, opened 2021-04-06
feat: Add Mypy linter

dhruvmanila:
Hello, awesome plugin. I was messing around with it and added the two linters: `mypy` and `flake8` for Python files. I will submit another PR for `flake8`.

`mypy` does not have the support to take input from stdin and so we have to use pattern matching to extract the out information from the output. I added a helper function `split` which can help in that. I put it in the `util` module as I believe it could be useful for other linters as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ example on the `InsertLeave` or `TextChanged` events.
 - [Languagetool][5]
 - [Vale][8]
 - [ShellCheck][10]
+- [Mypy][11]
 
 
 ## Custom Linters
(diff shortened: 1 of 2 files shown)


lua/lint/util.lua
```diff
@@ -22,4 +22,14 @@ function M.offset_to_position(lines, offset)
 end
 
 
+function M.split(str, sep)
```
mfussenegger, line 25: There is already `vim.split` or `vim.fn.split`, I think the former behaves pretty much the same. Could that be used instead?

mfussenegger (commented):
Left a question regarding the `split`, otherwise it looks good. Thank you for the PR :+1:

lua/lint/util.lua
dhruvmanila: Ah, completely missed that. I tried looking up a `split` command but it did not show up. I have made the necessary changes to use the `vim.fn.split` function.

PR #8, opened 2021-04-06
Add HTML Tidy

nanotee:
This PR adds HTML Tidy to the list of available linters

Since Tidy only reports a line and a column number, I chose to leave an offset of 1 like in #7


mfussenegger (commented):
Looks good. Thank you for the PR :)

PR #10, opened 2021-04-16
Add revive linter for go

xenrox:
Revive (https://github.com/mgechev/revive) is a linter for `go` that is a drop-in replacement for `golint`, so this PR could easily be adapted for that.
I based it on the `flake8` linter since both produce the same format for lint messages.


mfussenegger:
Thanks, looks good :+1:

PR #11, opened 2021-04-16
Add staticcheck linter for go

xenrox:
This still has two problems:
1. Just like the `shellcheck` linter, there is the message "Linter exited with code 1" after linting, even though it seems to work correctly.
1. It currently fails, unless I hardcode the values for end line/character, with
> Linter definition must have a `cmd` set: "vim.lua:129: .../packer/start/nvim-lint/lua/lint/linters/staticcheck.lua:25: '<name>' expected near 'end'

I guess that has to do with the keyword `end` in the json output.

Edit: Fixed issue two by reading more lua documentation.

### DIFF
--- a/lua/lint/linters/staticcheck.lua
+++ b/lua/lint/linters/staticcheck.lua
@@ -1,0 +1,36 @@
+local severities = {
+  error = vim.lsp.protocol.DiagnosticSeverity.Error,
+  warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  ignored = vim.lsp.protocol.DiagnosticSeverity.Information,
+}
+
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thank you for the PR. Looks good to me. 

> Just like the shellcheck linter, there is the message "Linter exited with code 1" after linting, even though it seems to work correctly.

I think we could add a `ignore_exitcode` property and if set skip the message.
Do you want to have a go at this?

xenrox:
> I think we could add a `ignore_exitcode` property and if set skip the message.
> Do you want to have a go at this?

Yes, I had a go at it.

PR #14, opened 2021-04-16
Ensure consistent Mypy output regardless of project configuration

lithammer:
This ensure a consistent output from Mypy regardless of what might be in a project's `mypy.ini` (or similar) that could alter the output format causing parsing errors.

It also removes the need for removing the "found n errors... / Success: no issues ..." message because it's not printed to begin with.


mfussenegger (approved):
Thanks :+1:

PR #15, opened 2021-04-16
Ensure consistent Flake8 output formatting

lithammer:
Similar to GH-14. Without this the output can be changed by local configuration files causing output parsing to fail.


mfussenegger:
:+1:

PR #16, opened 2021-04-17
Add pylint

xenrox:
Pylint returns a column of 0 for convention lint messages. For this the
substraction of 1 was removed so that using
`vim.lsp.diagnostic.goto_next` causes no problems.


mfussenegger:
Thank you, looks good :+1:

PR #17, opened 2021-04-24
Add golangci-lint linter

mekpavit:
Add `golangci-lint` linter. [golangci-lint](https://golangci-lint.run/) is a linter for Go that aggregate other popular Go linter together in one tool.


mfussenegger:
Looks good. thanks :+1:

PR #19, opened 2021-05-14
Add linter for Ruby

YorickPeterse:
This linter uses Ruby's built-in syntax checking option. In addition, we
make this linter the default linter to run for Ruby files.


mfussenegger (approved):
Thank you for the PR :+1:

PR #20, opened 2021-05-14
Add support for dynamic linter commands, and add the RuboCop linter

YorickPeterse:
This adds support for defining dynamic linter commands. In addition, support is added for the RuboCop linter; making use of the new dynamic command support.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -103,8 +103,31 @@ function M.lint(linter, client_id)
     cwd = vim.fn.getcwd(),
     detached = true
   }
-  assert(linter.cmd, 'Linter definition must have a `cmd` set: ' .. vim.inspect(linter))
-  handle, pid_or_err = uv.spawn(linter.cmd, opts, function(code)
+  local cmd = linter.cmd
+  if type(cmd) == 'function' then
+    cmd = cmd(bufnr)
+  end
+  -- When the function returns a table, the first value becomes the new
+  -- command. All other values are added to the start of the arguments list.
+  -- This makes it possible to dynamically switch between using for example
+  -- `rubocop` and `bundle exec rubocop` as the command.
+  if type(cmd) == 'table' then
+    local new_args = {}
+    local new_cmd = table.remove(cmd, 1)
+
+    for _, arg in ipairs(cmd) do
+      table.insert(new_args, arg)
+    end
+
+    for _, arg in ipairs(opts.args) do
+      table.insert(new_args, arg)
+    end
+
+    cmd = new_cmd
+    opts.args = new_args
+  end
+  assert(cmd, 'Linter definition must have a `cmd` set: ' .. vim.inspect(linter))
+  handle, pid_or_err = uv.spawn(cmd, opts, function(code)
     stdout:close()
     stderr:close()
     handle:close()
@@ -112,7 +135,7 @@ function M.lint(linter, client_id)
       print('Linter exited with code', code)
     end
   end)
-  assert(handle, 'Error running ' .. linter.cmd .. ': ' .. pid_or_err)
+  assert(handle, 'Error running ' .. cmd .. ': ' .. pid_or_err)
   local parser = linter.parser
   if type(parser) == 'function' then
     parser = require('lint.parser').accumulate_chunks(parser)
--- a/lua/lint/linters/rubocop.lua
+++ b/lua/lint/linters/rubocop.lua
@@ -1,0 +1,70 @@
+local severities = {
+  error = vim.lsp.protocol.DiagnosticSeverity.Error,
+  fatal = vim.lsp.protocol.DiagnosticSeverity.Error,
+  warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  refactor = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  convention = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  info = vim.lsp.protocol.DiagnosticSeverity.Warning,
+}
+
+local function use_bundler()
+  local dir = vim.fn.expand('%:p:h')
+
+  if vim.fn.findfile('Gemfile', dir .. ';') == '' then
+    return false
+  else
+    return true
+  end
+end
+
+return {
+  cmd = function(bufnr)
+    if use_bundler() then
+      return { 'bundle', 'exec', 'rubocop' }
+    else
+      return 'rubocop'
+    end
+  end,
+  stdin = false,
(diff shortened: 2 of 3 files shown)


YorickPeterse:
One alternative approach to dynamic commands is to unify `cmd` and `args` into a single field, then require that to either be a function that returns a list, or a list itself. This is actually a bit nicer to work with for linters with a weird CLI. 

For example, `gitlint` requires you to run `gitlint --config FILE lint` if you want to specify a custom config file. Running `gitlint lint --config ...` produces an error. You can work around this using the dynamic `cmd` feature added in this PR, but it's not quite what it was meant for.

This setup would result in linters being defined (more or less) like so:

```lua
return {
  cmd = { 'ruby', '-c', '-w' },
  ...
}
```

or this:

```lua
return {
  cmd = function(bufnr)
    return { 'ruby', '-c', '-w' }
  end,
  ...
}
```

lua/lint.lua
```diff
@@ -103,16 +103,39 @@ function M.lint(linter, client_id)
     cwd = vim.fn.getcwd(),
     detached = true
   }
-  assert(linter.cmd, 'Linter definition must have a `cmd` set: ' .. vim.inspect(linter))
-  handle, pid_or_err = uv.spawn(linter.cmd, opts, function(code)
+  local cmd = linter.cmd
+  if type(cmd) == 'function' then
+    cmd = cmd(bufnr)
```
YorickPeterse, line 108: On second thought, I have doubts about the use of passing the buffer number. I can't think of a case where you'd need to act on anything but the current buffer, if you even care about the explicit buffer number to begin with.

lua/lint.lua
```diff
@@ -103,16 +103,39 @@ function M.lint(linter, client_id)
...
+
+    for _, arg in ipairs(opts.args) do
+      table.insert(new_args, arg)
+    end
+
+    cmd = new_cmd
+    opts.args = new_args
+  end
+  assert(cmd, 'Linter definition must have a `cmd` set: ' .. vim.inspect(linter))
+  handle, pid_or_err = uv.spawn(cmd, opts, function(code)
```
mfussenegger, line 130: How about returning a tuple from the cmd function? The first return value is always a string, the second return value is optional and a list

`cmd, args = cmd()` 

Would simplfy this quite a bit.

What I'm unsure is if the `args` here are then an override or an extension of whatever is in the `linter.args` property.

lua/lint.lua
mfussenegger: Yes I also think it's not necessary. It can't be asynchronous, so the cmd function can always call `api.nvim_get_current_buf()` itself if it needs the current buffer number.

lua/lint.lua
YorickPeterse: @mfussenegger Ah, I forgot Lua supports multiple return values. 

The goal of the arguments here was to act as an extension of `linter.args`. But the more I think of it, the more I think it makes more sense to unify both into a single function (as I suggested in https://github.com/mfussenegger/nvim-lint/pull/20#issuecomment-841384109). What do you think of that approach?

YorickPeterse:
For linters such as `gitlint`, unifying the `cmd` and `args` settings may be required. While `gitlint` supports reading from STDIN, I can't get it to work through nvim-lint; for some reason gitlint just doesn't see the STDIN data. It does work with shellcheck though, so this seems to be a gitlint issue.

The problem with gitlint using a file as input is that you have to use `gitlint --msg-filename X lint`, using `gitlint --msg-filename lint X` or `gitlint lint --msg-filename X` won't work.

lua/lint/linters/rubocop.lua
```diff
@@ -0,0 +1,70 @@
+local severities = {
+  error = vim.lsp.protocol.DiagnosticSeverity.Error,
+  fatal = vim.lsp.protocol.DiagnosticSeverity.Error,
+  warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  refactor = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  convention = vim.lsp.protocol.DiagnosticSeverity.Warning,
+  info = vim.lsp.protocol.DiagnosticSeverity.Warning,
```
YorickPeterse, line 7: The choice for this mapping matches that ALE does. We could map it to other LSP severities such as Hint, but this may result in diagnostics not showing up when you have those filtered out. For example, I myself filter out hints and info messages because most language servers just produce a ton of noise using those diagnostic levels.

This behaviour also matches the rubocop CLI, in that the CLI doesn't care about the severity and treats all of them as offenses.

lua/lint.lua
mfussenegger: It sounds like the `gitlint` case could be solved by supporting `args` as a function? In both scenarios the `cmd` is `gitlint`.

I'm not to keen on conflating both into the `args` just to then split it apart again.

YorickPeterse: @mfussenegger I don't think that approach would work too well. For `gitlint` it would, because the base command is always `gitlint`. But for RuboCop (and possibly other linters) it means we have to repeat the same logic (a check to see if `bundle exec rubocop` is needed instead of just `rubocop`) in the `cmd` and `args` function. So something like this:

```lua
cmd = function()
  if use_bundler()
    return 'bundle'
  else
    return 'rubocop'
  end
end,
args = function(path)
  if use_bundler()
    return { 'exec', 'rubocop', path, ... }
  else
    return { path, ... }
  end
end
```

Unifying the two somehow means only a single check is necessary, and no repeated/redundant work needs to be performed.

mfussenegger: What would you think of supporting having the whole linter being a function? Instead of returning the linter table object, a linter can return a function that if called returns the linter table. So all options can be dynamically created if desired.

mfussenegger:
@YorickPeterse are you still interested in getting this in?

YorickPeterse:
@mfussenegger No, I have since switched to a custom linting solution, so feel free to close this PR :smiley:

jonatan-branting:
I would be interested in getting this merged. What still needs to be done before this can happen?

I'd be willing to finish the PR.

mfussenegger:
> I would be interested in getting this merged. What still needs to be done before this can happen?
> 
> I'd be willing to finish the PR.

My concerns were mostly in regards to the `cmd` / `args` framework changes.
The PR would need to be adapted to use the existing possibilities - e.g. return the whole linter as a function - similar to the [compiler.lua](https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/linters/compiler.lua) linter.

PR #32, opened 2021-06-21
Fixed golang revive linters matching pattern

coltonmorris:
`./path/to/file/main.go:15:2: don't use ALL_CAPS in Go names; use CamelCase` This error was causing the pattern matching for revive to fail.


mfussenegger:
Thanks, merged :+1:

PR #33, opened 2021-06-22
Feature: Add cppcheck support

wbthomason:
This PR adds initial, somewhat hacky support for [cppcheck](http://cppcheck.sourceforge.net/).

The hacky parts are:
1. It currently hardcodes the presence of a build directory named `build` and a `compile_commands.json` in that directory.
2. It runs `cppcheck` on the whole project and filters out diagnostics relevant to the current file.
3. It probably doesn't work right if a file has been modified but not saved.

This is very similar to the approach `ALE` takes, but I think we could do better by (1) more robust `compile_commands.json` detection and (2) maybe somehow parsing compile args from `compile_commands.json` ourselves to allow only running `cppcheck` on the currently relevant file (the tool is obnoxious and will either run on the whole project with compile args from a `compile_commands.json` file, or specific files with manually specified args).

Still, this has been a useful starting point for my use, so it could be worth adding in its current form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Other dedicated linters that are built-in are:
 | [Golangci-lint][16] | `golangcilint`  |
 | Ruby                | `ruby`          |
 | [Inko][17]          | `inko`          |
+| cppcheck            | `cppcheck`      |
 
 
 ## Custom Linters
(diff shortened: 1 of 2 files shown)


mfussenegger:
> Still, this has been a useful starting point for my use, so it could be worth adding in its current form.

Sounds good to me :+1:

PR #36, opened 2021-07-13
Add luacheck support

Dophin2009:
This PR adds [luacheck](https://github.com/mpeterv/luacheck) as a builtin linter for Lua.

### DIFF
--- a/lua/lint/linters/luacheck.lua
+++ b/lua/lint/linters/luacheck.lua
@@ -1,0 +1,41 @@
+local severities = {
+    W = vim.lsp.protocol.DiagnosticSeverity.Warning,
+    E = vim.lsp.protocol.DiagnosticSeverity.Error
+}
+
+local pattern = '[^:]+:(%d+):(%d+)-(%d+): %((%a)(%d+)%) (.*)'
+
+return {
+    cmd = 'luacheck',
+    stdin = true,
+    args = {'--formatter', 'plain', '--codes', '--ranges', '-'},
+    ignore_exitcode = true,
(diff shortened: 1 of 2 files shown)


lua/lint/linters/luacheck.lua
```diff
@@ -0,0 +1,40 @@
...
+    W = vim.lsp.protocol.DiagnosticSeverity.Warning,
+    E = vim.lsp.protocol.DiagnosticSeverity.Error
+}
+
+local pattern = '[^:]+:(%d+):(%d+)-(%d+): %((%a)(%d+)%) (.*)'
+
+return {
+    cmd = 'luacheck',
+    stdin = true,
+    args = {'--formatter', 'plain', '--codes', '--ranges', '-'},
```
mfussenegger, line 11: I think this also needs a `ignore_exitcode = false,` to suppress the "Linter command luacheck exited with code 1" messages.

mfussenegger:
Thanks for the PR :+1:

lua/lint/linters/luacheck.lua
Dophin2009: Shouldn't it be `ignore_exitcode = true`? I'm still getting the messages with `false` but they're gone with `true`.

mfussenegger: Right, it should be `true`

Dophin2009: Fixed.

Dophin2009:
Happy to help :+1:

PR #37, opened 2021-07-14
Add chktex linter

Dophin2009:
This PR adds [chktex](https://www.nongnu.org/chktex) as a builtin linter for TeX files.


mfussenegger (approved):
lgtm, thanks again for the PR

PR #39, opened 2021-07-15
luacheck: add vim global when checking runtime files

gpanders:
If the current buffer is modifying a Lua file found underneath an entry
in &runtimepath then add a read-only global 'vim'.

This suppresses the "undefined value" warnings in Lua files used from
Neovim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,20 @@ require('lint').linters.your_linter_name = {
 
 Instead of declaring the linter as a table, you can also declare it as a
 function which returns the linter table in case you want to dynamically
-generate some of the properties.
+generate some of the properties. This function is passed the buffer number of
+the buffer being linted:
+
+```lua
+return function(bufnr)
+  local cmd = ...; -- set command based on current buffer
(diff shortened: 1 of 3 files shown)


mfussenegger:
Shouldn't this rather be configured via a .luacheckrc file?

I understand that nvim-lint/luacheck will often be used to lint neovim plugin code,  but it feels wrong to me to hardcode project specific exceptions.

gpanders:
> On Jul 16, 2021, at 3:22 AM, Mathias Fußenegger ***@***.***> wrote:
> 
> ﻿
> Shouldn't this rather be configured via a .luacheckrc file?
> 
> I understand that nvim-lint/luacheck will often be used to lint neovim plugin code, but it feels wrong to me to hardcode project specific exceptions.
> 

Yes I think you’re right. I had this idea and perhaps got a little overexcited in wanting to upstream it right away :D But I agree, this should either go in a luacheckrc file or I can just redefine the luacheck linter in my own configuration rather than applying ad hoc rules for all users.

PR #41, opened 2021-07-17
Refactor from_pattern

Tastyep:
Hello,

I'd like to propose a change to the method named `from_pattern`, because it has a big impact I only recoded the function and will wait for your approval before going on.

The motivation is to support all the [Diagnostic](https://github.com/mfussenegger/nvim-lint#from_pattern) attributes and to make the parser flexible regarding what groups will be parsed using the regex.
This also allows adding new diagnostic fields in the future without worrying about refactoring.
Finally, I like the more concise and explicit definition of the parsers, for instance, I tested with cppcheck which could be defined as follows:
```
local pattern = [[([^:]*):(%d*):(%d*): %[([^%]\]*)%] ([^:]*): (.*)]]
local groups = {"file", "lineno", "colno", "code", "severity", "message" }
local severity_map = {
  ["error"] = vim.lsp.protocol.DiagnosticSeverity.Error,
  ["warning"] = vim.lsp.protocol.DiagnosticSeverity.Warning,
  ["performance"] = vim.lsp.protocol.DiagnosticSeverity.Warning,
  ["style"] = vim.lsp.protocol.DiagnosticSeverity.Information,
  ["information"] = vim.lsp.protocol.DiagnosticSeverity.Information,
}
local parser = from_pattern(pattern, groups, severity_map)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ The function allows to parse the linter's output using a lua regex pattern.
 - groups: The groups specified by the pattern
 
 ``` lua
-groups = {"lineno", "message", ("colno" | ("colbeg", "colend")), ["code"], ["codeDesc"], ["file"], ["severity"]}
+groups = {"line", "message", "start_col", ["end_col"], ["code"], ["code_desc"], ["file"], ["severity"]}
 ```
 
 - severity: A mapping from severity codes to diagnostic codes
--- a/README.md
+++ b/README.md
@@ -127,15 +127,34 @@ The function takes a single argument which is the `errorformat`.
 ### from_pattern
 
 ```lua
-parser = require('lint.parser').from_pattern(pattern, skeleton)
+parser = require('lint.parser').from_pattern(pattern, groups, severity_map, defaults)
 ```
 
-The function takes two arguments. The first is a lua pattern that must match
-four groups in order: line number, offset, code and message.
+The function allows to parse the linter's output using a lua regex pattern.
 
-The second argument is a skeleton used to create each diagnostic. This can be
-used to provide default values - for example to set a `severity`.
+- pattern: The regex pattern applied on each line of the output
+- groups: The groups specified by the pattern
 
+``` lua
+groups = {"lineno", "message", ("colno" | ("colbeg", "colend")), ["code"], ["codeDesc"], ["file"], ["severity"]}
+```
+
+- severity: A mapping from severity codes to diagnostic codes
+
+``` lua
+default_severity = {
+['error'] = vim.lsp.protocol.DiagnosticSeverity.Error,
+['warning'] = vim.lsp.protocol.DiagnosticSeverity.Warning,
+['information'] = vim.lsp.protocol.DiagnosticSeverity.Information,
+['hint'] = vim.lsp.protocol.DiagnosticSeverity.Hint,
+}
+```
+
+- defaults: The defaults diagnostic values
+
+``` lua
+defaults = {["source"] = "mylint-name"}
+```
 
 <details>
   <summary>Diagnostic interface description</summary>
--- a/lua/lint/linters/chktex.lua
+++ b/lua/lint/linters/chktex.lua
@@ -20,7 +20,7 @@ return {
             off = tonumber(off or 1) - 1
             d = tonumber(d or 1)
             table.insert(diagnostics, {
-                source = 'luacheck',
+                source = 'chktex',
                 range = {
                     ['start'] = {line = lineno, character = off},
                     ['end'] = {line = lineno, character = off + d}
--- a/lua/lint/parser.lua
+++ b/lua/lint/parser.lua
@@ -43,10 +43,6 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
       return entries['code_desc']
     end,
 
-    data = function(_)
-      return nil -- Unsupported
-    end,
-
     message = function(entries)
       return entries['message']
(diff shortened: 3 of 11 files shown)


mfussenegger (commented):
I like the idea :+1:

Tastyep:
I think I'm done with the parser refactoring.
I couldn't test `tidy` and `ruby`, it would be great if you can, I can confirm that the other ones I've modified are working.
I'll now update the [from_pattern](https://github.com/mfussenegger/nvim-lint#from_pattern) section of the README.

Tastyep:
Also, I've removed some logic preventing mypy to not follow imports. While I agree that it makes it verbose, since mypy allows this to be configurable, I don't think that nvim-lint should take responsibility for this.

Tastyep:
I think this is ready.

Regarding my point for mypy, the `file` group is now used to ensure that diagnostics from other files will be filtered out (which was the behavior implemented by some parsers (cppcheck, clang-tidy, mypy)).

README.md
```diff
@@ -127,15 +127,34 @@ The function takes a single argument which is the `errorformat`.
...
-four groups in order: line number, offset, code and message.
+The function allows to parse the linter's output using a lua regex pattern.
 
-The second argument is a skeleton used to create each diagnostic. This can be
-used to provide default values - for example to set a `severity`.
+- pattern: The regex pattern applied on each line of the output
+- groups: The groups specified by the pattern
 
+``` lua
+groups = {"lineno", "message", ("colno" | ("colbeg", "colend")), ["code"], ["codeDesc"], ["file"], ["severity"]}
```
mfussenegger, line 139:
```suggestion
groups = {"lineno", "message",  "start_col", ["end_col"], ["code"], ["codeDesc"], ["file"], ["severity"]}
```

`start_col` and `end_col` is what's typically used within the neovim code base. I also wouldn't distinguish between `colno` and `start_col`, but just have the former, and default `end_col` to `start_col` if it isn't present.

lua/lint/linters/chktex.lua
```diff
@@ -1,37 +1,35 @@
 local pattern = '(%d+):(%d+):(%d+):(.+):(%d+):(.*)'
 local severities = {
-    Error = vim.lsp.protocol.DiagnosticSeverity.Error,
-    Warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
-    Message = vim.lsp.protocol.DiagnosticSeverity.Information
+  Error = vim.lsp.protocol.DiagnosticSeverity.Error,
+  Warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
```
mfussenegger, line 4: Please revert the unrelated formatting changes

lua/lint/parser.lua
```diff
@@ -22,53 +21,108 @@ function M.from_errorformat(efm, skeleton)
...
+      end
+
+      local colno = entries['colno'] ~= nil and tonumber(entries['colno']) or 1
+      return {
+        ['start'] = { line = lineno - 1, character = colno - 1 },
+        ['end'] = { line = lineno - 1, character = colno },
+      }
+    end,
+
+    relatedInformation = function(_)
```
mfussenegger, line 71: Why do we already need methods for stuff that is not supported?

lua/lint/linters/chktex.lua
Tastyep: Sorry about that, I made a small change to the file and reformatted everything.
Do you currently use a formatter so that I can adapt to the same style ?

lua/lint/parser.lua
Tastyep: It was just to show how it should be implemented. Actually, I could just return entries[...] and it would be implemented, but as it was not used I didn't bother :).
Should I has support since it's easy to do ?

lua/lint/linters/chktex.lua
mfussenegger: I tend to use the formatting that you would've applied, but so far it is not enforced which is why different parsers have a different style. I wouldn't mind the reformat itself - what I dislike is that it reformats *and* makes a change within the same commit.

lua/lint/parser.lua
mfussenegger: I'd remove it if it is currently not used by any parser. We can extend it once needed.

Tastyep:
Thank you for the review, it's indeed better :+1:

mfussenegger:
Thanks again for the PR :+1:

lua/lint/parser.lua
```diff
@@ -22,53 +21,85 @@ function M.from_errorformat(efm, skeleton)
...
+      ['information'] = vim.lsp.protocol.DiagnosticSeverity.Information,
+      ['hint'] = vim.lsp.protocol.DiagnosticSeverity.Hint,
+    }
+  defaults = defaults or {}
+  return function(output, bufnr)
     local diagnostics = {}
+    local buffer_path = vim.api.nvim_buf_get_name(bufnr)
+
+    for _, line in ipairs(vim.fn.split(output, '\n')) do
+      local results = { line:gmatch(pattern)() }
```
gpanders, line 78: Out of curiosity, is there a reason to use `line:gmatch(pattern)()` rather than just `line:match(pattern)`? They do seem to be different, as with the former (current) implementation I am not getting any results from my [nix](https://github.com/mfussenegger/nvim-lint/pull/44) linter, but with the latter implementation it does work.

PR #47, opened 2021-07-20
Add clj-kondo for Clojure(script)

lispyclouds:
- Adds [clj-kondo](https://github.com/clj-kondo/clj-kondo) as a linter for clojure and clojurescript files


mfussenegger:
Thanks for the PR :+1:

PR #50, opened 2021-07-20
Add StandardRB

will:
This adds the ruby linter for standardrb.

This ends up looking like this
image

I was hoping that by returning the code that it would show up somehow, but I don’t see how to get at it. I was wondering if I should prepend it like this instead?

image

It's useful to know the name you want to disable that check.


mfussenegger:
Having the code or codeDescription show up would require a customization on the vim.lsp.diagnostic display side.

I think the PR is good as it is :+1:

PR #51, opened 2021-07-21
Improve clj-kondo setup

lispyclouds:
Based on feedback from the author @borkdude:
- Removed the --parallel option as it performs better on multiple files
  than one buffer which is the case here.
- Pass the file name from the buffer to help clj-kondo determine the
  type(clj, cljs, cljc, edn) better

### DIFF
--- a/lua/lint/linters/clj-kondo.lua
+++ b/lua/lint/linters/clj-kondo.lua
@@ -3,13 +3,17 @@ local severities = {
   warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
 }
 
+local function get_file_name(bufnr)
+  return vim.api.nvim_buf_get_name(bufnr)
+end
+
 return {
   cmd = 'clj-kondo',
   stdin = true,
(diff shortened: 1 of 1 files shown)


lua/lint/linters/clj-kondo.lua
```diff
@@ -3,13 +3,17 @@ local severities = {
   warning = vim.lsp.protocol.DiagnosticSeverity.Warning,
 }
 
+local function get_file_name(bufnr)
+  return vim.api.nvim_buf_get_name(bufnr)
+end
```
mfussenegger, line 8:
```suggestion
local function get_file_name()
  return vim.api.nvim_buf_get_name(0)
end
```

Currently no arguments are passed to the functions used within `args`.

lispyclouds: Right, making the change, i was trying to follow https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/linters/vale.lua#L2

PR #53, opened 2021-07-22
fix: use values from the defaults table

dhruvmanila:
The defaults table was not used at all because the loop would iterate
over only those keys present in the `group_handler` table.

An example would be `flake8` where the defaults contain the `source` key
but it is ignored as that key is absent from the `group_handler` table.

Solution:
Merge the diagnostic table with the defaults table after constructing
the diagnostic table as per the linter output.


mfussenegger:
Thanks, looks good to me :+1:

PR #58, opened 2021-07-24
fix: compare path relative to project root

dhruvmanila:
Linters output path relative to the project root and so we should
compare against that.

I noticed this with `mypy` where it would not send the diagnostics to the client as the path would differ.

Edit: Also, `mypy` exits with code 1, so I think we should ignore it by default similar to `flake8`

### DIFF
--- a/lua/lint/parser.lua
+++ b/lua/lint/parser.lua
@@ -72,7 +72,7 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
   defaults = defaults or {}
   return function(output, bufnr)
     local diagnostics = {}
-    local buffer_path = vim.api.nvim_buf_get_name(bufnr)
+    local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":~:.")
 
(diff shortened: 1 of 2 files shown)


lua/lint/parser.lua
```diff
@@ -72,7 +72,7 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
   defaults = defaults or {}
   return function(output, bufnr)
     local diagnostics = {}
-    local buffer_path = vim.api.nvim_buf_get_name(bufnr)
+    local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":~:.")
```
mfussenegger, line 75: Sure that all linters do this? I think it would be safer to also apply the `fnamemodify` call further down in the `entries['file'] == buffer_path` - that way it should also work if the linter reports the absolute path.

dhruvmanila: Ah, right. That makes sense.

mfussenegger:
Thanks :+1:

PR #64, opened 2021-07-29
flake8: use Warning as default severity

gpanders:
Since the overwhelming majority (maybe all?) of the "errors" that flake8 emits are, in fact, style/lint warnings, it makes more sense to default to using Warning as the default severity rather than Error.

### DIFF
--- a/lua/lint/linters/flake8.lua
+++ b/lua/lint/linters/flake8.lua
@@ -12,5 +12,6 @@ return {
   ignore_exitcode = true,
   parser = require('lint.parser').from_pattern(pattern, groups, nil, {
     ['source'] = 'flake8',
+    ['severity'] = vim.lsp.protocol.DiagnosticSeverity.Warning,
   }),
 }


dhruvmanila:
I believe there are two class of severity that flake8 reports: (W)arning and (E)rror as the code suggests and the rest can fallback to Info level. Is it possible to use the code to improve the reported severity?

gpanders:
The code doesn’t correspond to severity: see https://flake8.pycqa.org/en/latest/user/error-codes.html. Instead it’s more like the “class” of the “error” (“F” means it’s from flake8 directly, “W” is a white space error, “C” is a complexity error, etc). flake8 calls all of these things “errors”, but none of them match what I imagine the vast majority of people would consider an error in the context of a diagnostic (e.g. it prevents my code from compiling or running). For instance, “Line too long” or “extra whitespace before (“ are not errors in the usual sense, but warnings.

mfussenegger:
Makes sense to me. Thanks for the PR :+1:

PR #72, opened 2021-08-13
Add credo

zoedsoupe:
[Credo](https://github.com/rrrene/credo) is a static code analysis tool to check `Elixir` source code.

### DIFF
--- a/lua/lint/linters/credo.lua
+++ b/lua/lint/linters/credo.lua
@@ -1,8 +1,8 @@
 return {
-  cmd = 'mix credo',
-  args = { '--strict' },
+  cmd = 'mix',
+  args = { 'credo', '--strict' },
   ignore_exitcode = false,
-  parser = require('lint-parser').from_errorformat('%f:%l: %m', {
+  parser = require('lint.parser').from_errorformat('%f:%l: %m', {
     source = 'mix credo',
     severity = vim.lsp.protocol.DiagnosticSeverity.Information
   })
--- a/lua/lint/linters/credo.lua
+++ b/lua/lint/linters/credo.lua
@@ -1,0 +1,9 @@
+return {
+  cmd = 'mix credo',
+  args = { '--strict' },
+  ignore_exitcode = false,
+  parser = require('lint-parser').from_errorformat('%f:%l: %m', {
+    source = 'mix credo',
+    severity = vim.lsp.protocol.DiagnosticSeverity.Information
+  })
+}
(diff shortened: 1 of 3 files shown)


lua/lint/linters/credo.lua
```diff
@@ -0,0 +1,9 @@
+return {
+  cmd = 'mix credo',
+  args = { '--strict' },
```
mfussenegger, line 3:
```suggestion
  cmd = 'mix',
  args = { 'credo', '--strict' },
```

`cmd` must be a single command

lua/lint/linters/credo.lua
```diff
@@ -0,0 +1,9 @@
...
+  ignore_exitcode = false,
+  parser = require('lint-parser').from_errorformat('%f:%l: %m', {
```
mfussenegger, line 5:
```suggestion
  parser = require('lint.parser').from_errorformat('%f:%l: %m', {
```

mfussenegger (commented):
Thanks for the PR.

Left some suggestion because it looks as if it doesn't work - the tests also failed and I think the error format is wrong given the output.

matdsoupe:
Thanks for your review! I fixed the minor errors on parser definition but I didn't solve how to properly use the `error format`. I'm trying to solve this now!.

Also, I'm having trouble to run tests locally :/
I only need to clone plenary into somewhere and then put this repo on it?

mfussenegger:
> Also, I'm having trouble to run tests locally :/
> I only need to clone plenary into somewhere and then put this repo on it?


plenary needs to be a sibling to this repository. For example, if you keep all your plugins in `~/config/nvim/pack/plugins/start` you'd have:

```
~/.config/nvim/pack/plugins/start/plenary.nvim
~/.config/nvim/pack/plugins/start/nvim-lint
```

mfussenegger:
Are you still interested in finishing this up? Otherwise I'd close it

zoedsoupe:
Yeah, i'm in! i will update and do requested changes this week

skbolton:
I should have looked before I opened my PR - the elixir world is alive and well. #196 

Some comments on differences in our implementations where we can consolidate and pick whats best. There is a `--oneline` flag you can pass to `credo` that I think makes parsing a little easier. Also I see you have the `--strict` flag in yours. I went back and forth on this in mine and decided to go the safest route figuring people could override the args. But seeing yours makes me think its better to go strict and the user can always copy everything else.

mfussenegger:
Merged https://github.com/mfussenegger/nvim-lint/pull/196 instead

PR #74, opened 2021-08-22
Add selene

midnightexigent:
`selene` is a linter for lua

`null-ls` supports it so I added it to this plugin as well


mfussenegger:
Now you overshot a bit :)
Please don't reformat the entire README.

midnightexigent:
ha, interesting, I didn't even do that on purpose, that was prettier doing some format on save magic. 

how about now ?

mfussenegger:
Seems okay. Thanks for the PR

PR #85, opened 2021-09-01
Feature/add pycodestyle linter

jeroen-rooijmans:
Thanks for this plugin, it totally fits my needs! I've added pycodestyle linter included some basic tests, I prefer pycodestyle for linting over flake8, mypy or pylint as these all have more overlap with the python LSP. It's very much based to the flake8 linter that was already available.

### DIFF
--- a/tests/pycodestyle_spec.lua
+++ b/tests/pycodestyle_spec.lua
@@ -1,7 +1,0 @@
-describe('linter.pycodestyle', function()
-  it("doesn't error on empty output", function()
-    local parser = require('lint.linters.pycodestyle').parser
-    parser('')
-    parser('  ')
-  end)
-end)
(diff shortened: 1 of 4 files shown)


jeroen-rooijmans:
I've noticed that the pycodestyle output is not coming through. I'll close this for now and continue to work on it.

tests/pycodestyle_spec.lua
```diff
@@ -0,0 +1,7 @@
+describe('linter.pycodestyle', function()
+  it("doesn't error on empty output", function()
```
mfussenegger, line 2: Could you move this test case into the other file as well?

mfussenegger (commented):
Thanks for the PR. Left a minor comment regarding having the test cases across two files, otherwise lgtm

Issue #87, opened 2021-09-02
Cppcheck >1.84 required ({column} required)


tsjordan-eng:
Can we add a note in the readme that states `cppcheck` version 1.84 or greater is required? I went through a lot of debugging to find out that 1.82 that in Ubuntu's repo doesn't support `{column}` in the template argument. No indication was given by nvim-lint that cppcheck was failing.
 (Is there a good way to debug this tool and the parser? LspInfo doesn't show anything.)

tsjordan-eng:
I'm trying to solve this problem in #88

mfussenegger:
https://github.com/mfussenegger/nvim-lint/pull/88 got merged

PR #88, opened 2021-09-02
Add backwards compatibility to cppcheck 1.34

tsjordan-eng:
Allow parser to parse `{column}` or the actual column number. Default
column to 1.

Addresses #87

### DIFF
--- a/lua/lint/linters/cppcheck.lua
+++ b/lua/lint/linters/cppcheck.lua
@@ -1,4 +1,5 @@
-local pattern = [[([^:]*):(%d*):(%d*): %[([^%]\]*)%] ([^:]*): (.*)]]
+-- cppcheck <= 1.84 doesn't support {column} so the start_col group is ambiguous
+local pattern = [[([^:]*):(%d*):([^:]*): %[([^%]\]*)%] ([^:]*): (.*)]]
 local groups = { 'file', 'line', 'start_col', 'code', 'severity', 'message' }
 local severity_map = {
   ['error'] = vim.lsp.protocol.DiagnosticSeverity.Error,
--- a/lua/lint/parser.lua
+++ b/lua/lint/parser.lua
@@ -49,7 +49,7 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
 
     range = function(entries)
       local line = tonumber(entries['line'])
-      local start_col = tonumber(entries['start_col'])
+      local start_col = tonumber(entries['start_col']) or 1
       local end_col = entries['end_col'] and tonumber(entries['end_col']) or start_col
       return {
         ['start'] = { line = line - 1, character = start_col - 1 },
--- a/tests/cppcheck_spec.lua
+++ b/tests/cppcheck_spec.lua
@@ -1,16 +1,17 @@
 describe('linter.cppcheck', function()
   it('can parse the output', function()
     local parser = require('lint.linters.cppcheck').parser
+    local bufnr = vim.uri_to_bufnr('file:///foo.cpp')
     local result = parser([[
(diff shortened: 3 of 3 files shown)


tsjordan-eng:
I tried to make a test based on the Markdown one, but it failed on counting entries in `result` I'm afraid I don't know Lua well enough.
```lua
describe('linter.cppcheck', function()
  it('can parse the output', function()
    local parser = require('lint.linters.cppcheck').parser
    local result = parser([[
foo.cpp:46:7: [unusedVariable] style: Unused variable: fd
foo.cpp:366:3: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
foo.cpp:46:{column}: [unusedVariable] style: Unused variable: fd
foo.cpp:366:{column}: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
]])

  assert.are.same(4, #result)

  local expected = {
    source = 'cppcheck',
    message = 'Unused variable: fd',
    range = {
      ['start'] = {
        character = 6,
        line = 46,
      },
      ['end'] = {
        character = 7,
        line = 46,
      }
    },
    severity = vim.lsp.protocol.DiagnosticSeverity.Information,
  }
  assert.are.same(expected, result[1])

  local expected = {
    source = 'cppcheck',
    message = 'Unused variable: fd',
    range = {
      ['start'] = {
        character = 0,
        line = 46,
      },
      ['end'] = {
        character = 1,
        line = 46,
      }
    },
    severity = vim.lsp.protocol.DiagnosticSeverity.Information,
  }
  assert.are.same(expected, result[3])

  end)
end)
```

mfussenegger:
The test case you posted seems alright, if it doesn't pass - could it be due to the pattern not matching correctly?

tsjordan-eng:
The pattern matches and works in practice, just not in the test.

I found that I'm being blocked in `lint/parser.lua:88`. I'm not sure what the logic is supposed to be here:
If there's **not** a file name in the line *or* that file name doesn't match the buffer name? I don't know what the `not` is for.

Either way I see that this only makes the diagnostics if the matched file name is the name as the buffer name. Can we get around this for testing?

mfussenegger:
The following would work:

```
diff --git a/tests/cppcheck_spec.lua b/tests/cppcheck_spec.lua
index 5cc7219..f0b0108 100644
--- a/tests/cppcheck_spec.lua
+++ b/tests/cppcheck_spec.lua
@@ -1,14 +1,15 @@
 describe('linter.cppcheck', function()
   it('can parse the output', function()
     local parser = require('lint.linters.cppcheck').parser
+    local bufnr = vim.uri_to_bufnr('file:///foo.cpp')
     local result = parser([[
-./foo.cpp:46:7: [unusedVariable] style: Unused variable: fd
-./foo.cpp:366:3: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
-./foo.cpp:46:{column}: [unusedVariable] style: Unused variable: fd
-./foo.cpp:366:{column}: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
-]])
+/foo.cpp:46:7: [unusedVariable] style: Unused variable: fd
+/foo.cpp:366:3: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
+/foo.cpp:46:{column}: [unusedVariable] style: Unused variable: fd
+/foo.cpp:366:{column}: [postfixOperator] performance: Prefer prefix ++/-- operators for non-primitive types.
+]], bufnr)

   assert.are.same(4, #result)

   local expected = {
     source = 'cppcheck',
```

mfussenegger:
Looks great now. I merged it - thanks for adding the test cases :+1:

Issue #92, opened 2021-09-08
Diagnostic error messages cut off at punctuation marks


heygarrett:
Example LSP diagnostic errors from nvim-lint, using `vim.lsp.diagnostic.set_loclist()`:  

```
web/src/index.ts|5 col 5 error| Promises must be handled appropriately or explicitly marked as ignored with the
web/src/index.ts|15 col 5 error| Unsafe member access
web/src/index.ts|15 col 5 error| Unsafe call of an
web/src/index.ts|16 col 5 error| Unsafe member access
web/src/index.ts|16 col 5 error| Unsafe call of an
```

And the output from eslint run on the same file:  

```
web/src
❯ eslint index.ts

/path/to/web/src/index.ts
   5:5  error  Promises must be handled appropriately or explicitly marked as ignored with the `void` operator  @typescript-eslint/no-floating-promises
  15:5  error  Unsafe member access .accept on an `any` value                                                   @typescript-eslint/no-unsafe-member-access
  15:5  error  Unsafe call of an `any` typed value                                                              @typescript-eslint/no-unsafe-call
  16:5  error  Unsafe member access .dispose on an `any` value                                                  @typescript-eslint/no-unsafe-member-access
  16:5  error  Unsafe call of an `any` typed value                                                              @typescript-eslint/no-unsafe-call

✖ 5 problems (5 errors, 0 warnings)
```

It seems that error messages get cut off for both backticks and periods, though I'm not sure if other punctuation marks create the same behavior.

Issue #69, opened 2021-08-09
Eslint Setup Help


KasonBraley:
I am having trouble setting this plugin up so I can lint JS files with `eslint`. Would you be able to provide so examples or any pointers on how to get started? I don't see anywhere on the Readme about how the errors from the linter are displayed, is that handled by neovim's LSP instead or something else I would need to setup?

Here is my `nvim-lint` config. 
```
local lint = require "lint"

lint.linters_by_ft = {
  javascript = {"eslint"}
}

vim.cmd([[au BufEnter *.js lua require('lint').try_lint() ]])
vim.cmd([[au BufWritePost *.js lua require('lint').try_lint() ]])
```

Some things I have tried to get it to work:

- I have tried using eslint as both a global package as well as a dev dependency
- Explicitly specified elsint's args like mentioned in #31
- tried setting `lint.linters.eslint.cmd = "node_modules/.bin/eslint"` when i installed the eslint package as a dev dependency

Any help is appreciated!

mfussenegger:
Yes the diagnostics are reported via neovim's LSP diagnostic module. By default you should get signs displayed, highlights and in-line messages via virtual-text. You can customize that. See `:help vim.lsp.diagnostic.on_publish_diagnostics()`.

Your config looks okay.

> tried setting lint.linters.eslint.cmd = "node_modules/.bin/eslint" when i installed the eslint package as a dev dependency

I think the path needs to be absolute. 

You can try running `:lua require('lint').try_lint('eslint')` manually to see if it works or if there is an error.

mfussenegger:
> I did that and I received `Linter command ./node_modules/.bin/eslint exited with code 1`

Looks like the linter definition would need the `ignore_exitcode = true,` option.
Could you post some sample linter output in text form? I don't use eslint myself but that would allow me to create a test case.

heygarrett:
I'm also not seeing diagnostics from eslint.

Current config:  
```lua
local lint = require('lint')
lint.linters.eslint.cmd = './node_modules/.bin/eslint'
lint.linters_by_ft = {
	javascript = {'eslint'},
	typescript = {'eslint'}
}
vim.cmd([[au BufWritePost <buffer> lua require('lint').try_lint()]])
```

<img width="305" alt="Screen Shot 2021-09-01 at 18 37 29" src="https://user-images.githubusercontent.com/1154912/131754432-bb5f5397-50eb-4a7c-a4d1-8e616c16db53.png">

[...]

PR #93, opened 2021-09-09
More effective eslint capture groups

heygarrett:
Punctuation in error messages is now captured properly, which prevents some error messages from getting cut off or left out completely.

Addresses #69 and #92.


mfussenegger:
Thanks for the PR

PR #111, opened 2021-11-03
Add Statix

houstdav000:
Add the statix linting and suggestion tool for the Nix programming
language to both the linter configs and the list of available linters.

It uses errorformat, so it was really easy to write, and last I saw it
worked properly and found some issues with my dotfiles' flake.nix
file, defined manually and by using this fork.

Let me know if I should add a test or something. I'm just not sure
what I'd test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ Other dedicated linters that are built-in are:
 | [Selene][31]                 | `selene`       |
 | [ShellCheck][10]             | `shellcheck`   |
 | [StandardRB][27]             | `standardrb`   |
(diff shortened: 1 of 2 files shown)


mfussenegger:
Looks good,  thanks

> Let me know if I should add a test or something. I'm just not sure
what I'd test.

It can be helpful to test that some output is actually matched by a pattern or the errorformat, especially if the format isn't straightforward.

So far I don't require tests for new linters and leave it to contributors to judge if its warranted (unless I spot something that's obviously broken)

PR #112, opened 2021-11-05
Add mlint

mendes-davi:
`mlint` is a Matlab linter provided by Mathworks. I also provided a minimal test.

This my very first time using LUA besides my nvim init.lua so feel free to review the implementation.
Thanks!

### DIFF
--- a/lua/lint/linters/mlint.lua
+++ b/lua/lint/linters/mlint.lua
@@ -1,4 +1,5 @@
 local pattern = "L (%d+) %(C (%d+)[-]?([%d]*)%): (.-): (.-): (.*)"
+local groups = { "line", "start_col", "end_col", "code", "severity", "message" }
 local severities = {
   ML2 = vim.lsp.protocol.DiagnosticSeverity.Error,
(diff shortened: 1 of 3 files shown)


lua/lint/linters/mlint.lua
```diff
@@ -0,0 +1,43 @@
...
+  stdin = false,
+  stream = "stderr",
+  args = { "-cyc", "-id", "-severity" },
+  ignore_exitcode = true,
+  parser = function(output, _)
+    local result = vim.fn.split(output, "\n")
+    local diagnostics = {}
+
+    for _, line in ipairs(result) do
+      local lineno, start_col, end_col, code, sev, desc = string.match(line, pattern)
```
mfussenegger, line 22: Given that this uses `string.match` with a pattern, would it be possible to use the `from_pattern` function?

See `pycodestyle`, `mypy`, `eslint` or `clangtidy` for examples.

mendes-davi: I think it might be possible...

mendes-davi: Thanks! I ended up using the same logic as `from_pattern`.

PR #121, opened 2021-11-25
Add robocop

Yutsuten:
[Robocop](https://github.com/MarketSquare/robotframework-robocop) is a newer linter for the Robot Framework. Seems to be better than rflint.

PRing it for anyone else who needs it!


mfussenegger (commented):
Could you also add it to the linters table in the readme?

Yutsuten:
Added it to the linters table! By the way I fixed the link of rflint too.

PR #126, opened 2021-12-01
Fix #125, Fix incorrect testing rtp, and add pydocstyle linter config

### LINKED ISSUE #125: from_errorformat should strip leading newlines and whitespace from result of getqflist item text field
When neovim (or vim) parses lines of text with an `errorformat` string, where `errorformat` is mutli-line, the text matching `%m` can end up starting with a newline character regardless of the position of `%m` within the line, if it exists on a matching line that isn't the first one.

To see this, run the following in neovim/vim and note the output begins with a blank line:

    :echo getqflist({'efm': '%Afoo,%Zbar %m', 'lines': ['foo', 'bar baz']})[0]['text']

In other words, the 'text' field above is `'\nbaz'` rather than just the expected `'baz'`. AFAIK, this is a quirk of the internals of how `errorformat` string parsing works (if anyone knows differently please comment because I'd love a general solution to this vim problem).

When neovim/vim displays the quickfix list, there is code which explicitly strips leading newlines and spaces from the resultant message to account for this situation. See the `qf_fmt_text` function [here](https://github.com/neovim/neovim/blob/master/src/nvim/quickfix.c#L3166) and its use when adding new content to the quickfix buffer [here](https://github.com/neovim/neovim/blob/master/src/nvim/quickfix.c#L3992) and [here](https://github.com/neovim/neovim/blob/master/src/nvim/quickfix.c#L4003).

However, in this project, `from_errorformat()` calls `getqflist()` [here](https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/parser.lua#L17) to parse an `errorformat` string (which may contain the above situation) and the result is returned to be displayed as it is. The result is that any linter configuration created that has the above condition will wind up with a newline displayed before the error message text. So, for example, diagnostics might look like this:

    Diagnostics:
    1.
    Some error message
    2.
    Another error message

Consequently, I'd like to recommend `nvim-lint` maintain parity with the behavior of the quickfix list and strip leading newlines and spaces from error messages in `from_errorformat()`. It would probably also be a good idea to strip trailing whitespace at the same time while at it. Specifically, change [this line](https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/parser.lua#L31) to:

      message = item.text:match("^%s*(.-)%s*$"),

To work around this issue currently, one needs to intercept the message somewhere in the neovim config and strip it themselves. For example, I have my diagnostics display in floating windows and use version 0.6, so I do this:

    vim.diagnostic.config({
      float = {
        format = function(diag)
          return diag.message:match("^%s*(.-)%s*$")
        end
      }
    })

abbottc:
__NOTES:__

- The pydocstyle config has the multi-line errorformat situation described in #125 and therefore it will benefit from the included fix for it. 
- The rtp fix is something I noticed after a long while having the test suite failing doing this PR before discovering the problem in the test setup.
- See extended commit messages for specific detail on each commit

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Other dedicated linters that are built-in are:
 | [Mypy][11]                   | `mypy`         |
 | nix                          | `nix`          |
 | [pycodestyle][pcs-docs]      | `pycodestyle`  |
+| [pydocstyle][pydocstyle]     | `pydocstyle`   |
 | [Pylint][15]                 | `pylint`       |
 | [Revive][14]                 | `revive`       |
 | [rflint][rflint]             | `rflint`       |
@@ -321,6 +322,7 @@ nvim --headless --noplugin -u tests/minimal.vim -c "PlenaryBustedDirectory tests
 [plenary]: https://github.com/nvim-lua/plenary.nvim
 [ansible-lint]: https://docs.ansible.com/lint.html
 [pcs-docs]: https://pycodestyle.pycqa.org/en/latest/
+[pydocstyle]: https://www.pydocstyle.org/en/stable/
 [checkstyle]: https://checkstyle.sourceforge.io/
 [jshint]: https://jshint.com/
 [rflint]: https://github.com/boakley/robotframework-lint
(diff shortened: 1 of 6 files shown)


mfussenegger:
Nice :+1:

PR #140, opened 2021-12-22
selene further improvements

jurij-robba:
warning also doesn't work, so a mapping is required


mfussenegger (commented):
Looks good but there are conflicts, could you rebase against master?

jurij-robba:
should be good now

PR #142, opened 2021-12-27
Add vulture

mflova:
[Vulture](https://github.com/jendrikseipp/vulture) is the best linter I have found when it comes to finding dead code in Python.
This is my very firest PR in github and totally new to lua/regex, so just tell me if something is wrong or missing :)
Since the arguments of vulture are just a few, this implementation should be robust regardless its input arguments.

### DIFF
--- a/lua/lint/linters/vulture.lua
+++ b/lua/lint/linters/vulture.lua
@@ -1,0 +1,14 @@
+-- path/to/file:line:col: code message
+local pattern = '([^:]+):(%d+):(.*)'
+local groups = { 'file', 'lnum', 'message' }
+
+return {
+  cmd = 'vulture',
+  stdin = false,
+  args = {},
+  ignore_exitcode = true,
+  parser = require('lint.parser').from_pattern(pattern, groups, nil, {
+    ['source'] = 'vulture',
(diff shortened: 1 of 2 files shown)


lua/lint/linters/vulture.lua
```diff
@@ -0,0 +1,14 @@
+-- path/to/file:line:col: code message
+local pattern = '([^:]+):(%d+):(.*)'
```
mfussenegger, line 2: If the output is `path/to/file:line:col: code message`, shouldn't there be another pattern and group for the column?

mfussenegger (commented):
Thanks for the PR . Left a question around the pattern, otherwise this looks good as far as I can tell

lua/lint/linters/vulture.lua
mflova: This comment comes from another file I used as a template. An example output is the next one, where no column or code is displayed:
```
/home/user/test/test.py:2: unused variable 'i' (60% confidence)
/home/user/test/test.py:3: unused variable 'test_variable' (60% confidence)
```
I updated that comment just in case. I might be adding othger linters I tipically use. Thanks!

PR #154, opened 2022-01-04
fix: improve vulture scope

### LINKED ISSUE #153: Linters that operate at directory or repository level
Right now, one of the linters I added (`vulture`, the best linter I have found based on dead and unused variables) is adding lots of false positives.
It seems that this linter works perfectly when the full directory is provided as input, as it scans all the files where the variables might have been used. This can be the repository or even greater scope, to analyze which functions might have been used in other repositories in just less than a few seconds.
However, in `nvim-lint` you can only specify the filename when the parameter stdin is set to false, which limits the scope of some linters to analize out of the current file.

Is there any linter added that uses a greater scope than the filename and then filter all those errors to only ones that belong to the current file? I would like to implement a fix, but I am not sure where to start. I might try something in the meantime.

mflova:
The idea of this PR is to improve the scope of Vulture (dead code and unused variables analyzer). By default, the filename is passed as argument, which prevents the linter to analyze the full context of an entire repository. This leads to lots of false positives. The current implementation uses as input argument the git root directory if exists or the current dir otherwise. Tested both of them and works fine.

Fixes #153

### DIFF
--- a/lua/lint/linters/vulture.lua
+++ b/lua/lint/linters/vulture.lua
@@ -1,12 +1,22 @@
 -- path/to/file:line: message
-local pattern = '([^:]+):(%d+):(.*)'
+local pattern = '([^:]+):(%d+): (.*)'
 local groups = { 'file', 'lnum', 'message' }
 
 return {
   cmd = 'vulture',
   stdin = false,
-  args = {},
+  args = {"--exclude='/**/docs/*.py,/**/build/*.py'", function()
+    local output = vim.fn.system("git rev-parse --show-toplevel"):sub(1, -2)
+    if output:match("^([%w]+)") == "fatal" then
+        -- Return the current dir
+        return vim.fn.getcwd()
+    else
+        -- Return the path to git root directory
+        return output
+    end
+  end},
   ignore_exitcode = true,
+  append_fname = false,
   parser = require('lint.parser').from_pattern(pattern, groups, nil, {
     ['source'] = 'vulture',
     ['severity'] = vim.diagnostic.severity.WARN,
--- a/lua/lint/linters/vulture.lua
+++ b/lua/lint/linters/vulture.lua
@@ -6,13 +6,11 @@ return {
   cmd = 'vulture',
   stdin = false,
   args = {"--exclude='/**/docs/*.py,/**/build/*.py'", function()
-    local output = vim.fn.system("git rev-parse --show-toplevel"):sub(1, -2)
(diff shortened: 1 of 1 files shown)


mflova:
To be honest I do not really understand how is this issue in the tests related to this PR. In my PC is failing `nvim-lint/tests/compiler_spec.lua`, while here two random and compeltely different linters are failing the test. I even checkout previous commits from my PR that were OK and they are also failing locally.

lua/lint/linters/vulture.lua
```diff
@@ -1,12 +1,22 @@
...
-local pattern = '([^:]+):(%d+):(.*)'
+local pattern = '([^:]+):(%d+): (.*)'
 local groups = { 'file', 'lnum', 'message' }
 
 return {
   cmd = 'vulture',
   stdin = false,
-  args = {},
+  args = {"--exclude='/**/docs/*.py,/**/build/*.py'", function()
+    local output = vim.fn.system("git rev-parse --show-toplevel"):sub(1, -2)
```
mfussenegger, line 9: Why the `:sub(1, -2)` ?

lua/lint/linters/vulture.lua
```diff
@@ -1,12 +1,22 @@
...
+    if output:match("^([%w]+)") == "fatal" then
+        -- Return the current dir
+        return vim.fn.getcwd()
+    else
+        -- Return the path to git root directory
+        return output
+    end
+  end},
```
mfussenegger, line 17:
```suggestion
    local git_root_or_err = vim.fn.system("git rev-parse --show-toplevel"):sub(1, -2)
    if vim.startswith(git_root_or_err, 'fatal') then
        return vim.fn.getcwd()
    else
        return git_root_or_err
    end
  end},
```

mfussenegger:
> To be honest I do not really understand how is this issue in the tests related to this PR

The failures were realted to upstream changes in neovim itself, see https://github.com/mfussenegger/nvim-lint/pull/155
The tests currently run against neovim nightlies.

lua/lint/linters/vulture.lua
mflova: Because the command retrieves a string with a special character at the end. This character, when used as input argument to vulture, ended up in the linter not properly working (basically there were no diagnostics). I verified that the current resulting string is the correct one by 1) comparing it with a hardcoded string of the expected path and 2) observing the same diagnostics were produced in the Vulture CLI and in Neovim

PR #157, opened 2022-01-07
feat: add pytest-cov

mflova:
A long one here! This time for [pytest-cov](https://pypi.org/project/pytest-cov/), which launches pytest to check for test coverage of different files.
I recently saw that there were some implementations about test coverage in vim, so why not bringing then to nvim built-in diagnostics? The CLI output is the next one:
```bash
plugins: typeguard-2.13.3, cov-3.0.0, pylama-8.3.3, cases-3.6.7, harvest-1.10.3
collected 3 items                                                                                                                                                                                                 

tests/mymodule/test_additions.py ...                                                                                                                                                                        [100%]

---------- coverage: platform linux, python 3.8.10-final-0 -----------
Name                               Stmts   Miss  Cover   Missing
----------------------------------------------------------------
main.py                                9      9     0%   1-14
myapp/__init__.py                      0      0   100%
myapp/app.py                           5      5     0%   1-9
myapp/mymodule/__init__.py             0      0   100%
myapp/mymodule/funcs.py                4      1    75%   4
tests/__init__.py                      0      0   100%
tests/mymodule/__init__.py             0      0   100%
tests/mymodule/test_additions.py       6      0   100%
----------------------------------------------------------------
TOTAL                                 24     15    38%

```
Here is the output. If something is not clear (messages, severities...) it can be also changed of course:

When no test coverage is found for a given file:
pytestcov_notesting

When it is found, but some lines were not tested:
pytestcov_50

When the file could not be tested due to errors
pytestcov_errors

However, there are a few things related to it. Some of them as general comments and some of them as opinion:
- In case the user disables the underline, I added a new commit in which the message include "(until X line)" to know where the non-tested block ends.
- About the underline, it starts to be really annoying in large code scripts. I was thinking about disabling it by default (ie set the col diagnostics to 0). Opinions? :) 
- Since there is now way to know which tests test which scripts, `pytest` will launch all tests in the given directory.

### DIFF
--- a/lua/lint/linters/pytestcov.lua
+++ b/lua/lint/linters/pytestcov.lua
@@ -75,10 +75,10 @@ return {
 
     if tonumber(percentage_cover) == 0 then
       table.insert(diagnostics, {
-         lnum = 0,
-         col = 0,
-         severity = assert(vim.diagnostic.severity.INFO, 'missing mapping for severity info'),
-         message = messages.file_not_covered,
+        lnum = 0,
+        col = 0,
+        severity = assert(vim.diagnostic.severity.INFO, 'missing mapping for severity info'),
+        message = messages.file_not_covered,
       })
     else
       for non_covered_lines in non_covered_blocks:gmatch("([^,]+),?") do
--- a/lua/lint/linters/pytestcov.lua
+++ b/lua/lint/linters/pytestcov.lua
@@ -15,7 +15,10 @@ local ignored_files = {
   "docs/conf.py",
   "test_.*.py",
   "__init__.py",
+  "conftest.py",
 }
+
+local source_pytestcov = "pytest-cov"
 
 return {
   cmd = 'pytest',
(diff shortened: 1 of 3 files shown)


lua/lint/linters/pytestcov.lua
```diff
@@ -0,0 +1,106 @@
...
+  "setup.py",
+  "docs/conf.py",
+  "test_.*.py",
+  "__init__.py",
+}
+
+return {
+  cmd = 'pytest',
+  stdin = false,
+  args = {"--cov-reset", "--cov-report=term-missing", "--cov=."},
```
mflova, line 23: Would there be a way to retrieve the durrent directory of the buffer in the args? `--cov=.` would print the report for all the entire directory, when it is only needed the one where the buffer file is located

mfussenegger:
CI should be fixed. 

I'm not entirely sure if this still fits into nvim-lint and I'm kinda reluctant on merging it.
I'd keep it open for a while to see if there is more interest in this

PR #158, opened 2022-01-07
fix: better information retrieved pylint and codespell

mflova:
Just a small PR. Since I use Trouble to look at the diagnostics, I could not see some of the information that some linters are supposed to retrieve.
I guess that the previous implementation of Pylint did not read the `message-id` because it was not possible. It seems that, in order to read a value from a lua table, all the keys have to be valid identifiers (which means no hyphens are allowed). Due to this I replaced `message-id`by `message_id`. Thanks to it it was possible to be read.

There might be more linters without this info, but I do not want to modify those that I cannot test. These are the oens I verified:
```  
python = {'flake8', 'mypy', 'pylint', 'vulture', 'codespell', 'pytestcov'},
cpp = {'cppcheck', 'clangtidy', 'cpplint', 'codespell'},
rst = {'rstlint', 'rstcheck', 'proselint', 'codespell'},
markdown = {'markdownlint', 'codespell', 'proselint'},
yaml = {'yamllint', 'codespell', 'proselint'},
cmake = {'cmakelint'},
vim = {'vint'},
```

Before
pylint_before
After
pylint_after

### DIFF
--- a/lua/lint/linters/pylint.lua
+++ b/lua/lint/linters/pylint.lua
@@ -16,6 +16,7 @@ return {
     local diagnostics = {}
     local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":~:.")
 
+    output = string.gsub(output, "message%-id", "message_id")
     for _, item in ipairs(vim.fn.json_decode(output) or {}) do
       if not item.path or vim.fn.fnamemodify(item.path, ":~:.") == buffer_path then
         local column = 0
@@ -23,12 +24,14 @@ return {
           column = item.column - 1
         end
         table.insert(diagnostics, {
+          source = 'pylint',
           lnum = item.line - 1,
           col = column,
           end_lnum = item.line - 1,
           end_col = column,
           severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
           message = item.message,
+          code = item.message_id,
         })
       end
     end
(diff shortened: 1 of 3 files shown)


lua/lint/linters/pylint.lua
```diff
@@ -16,19 +16,22 @@ return {
...
         end
         table.insert(diagnostics, {
+          source = 'pylint',
           lnum = item.line - 1,
           col = column,
           end_lnum = item.line - 1,
           end_col = column,
           severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
           message = item.message,
+          code = item.message_id,
```
mfussenegger, line 34:
```suggestion
    for _, item in ipairs(vim.fn.json_decode(output) or {}) do
      if not item.path or vim.fn.fnamemodify(item.path, ":~:.") == buffer_path then
        local column = 0
        if item.column > 0 then
          column = item.column - 1
        end
        table.insert(diagnostics, {
          source = 'pylint',
          lnum = item.line - 1,
          col = column,
          end_lnum = item.line - 1,
          end_col = column,
          severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
          message = item.message,
          user_data = {
            lsp = {
              code = item['message-id'],
            },
          },
```

No need to replace the message id column in the string. table access is possible for words with hyphens as well.
And there is no first class `code` property (see `:help diagnostic-structure`). For LSP items it's put into the `user_data` property.

mflova: Oh, I have just seen the mapping you do in `from_pattern`. Somehow it was working. Fixed done! I also had to update the tests.

PR #165, opened 2022-01-16
Add phpcs linter

acburdine:
First - thanks for this plugin! Switched over from `ale` recently and am loving the speed this plugin has.

Was in need of a `phpcs` lint integration, so I wrote my own. This integration works locally, but this is my first time writing Lua so am happy to implement any suggestions 😄

Edit: Looks like the Github Actions won't run automatically, but I did run the tests I added locally and they did pass.


mfussenegger:
Looks good to me. :+1:

PR #179, opened 2022-02-04
feat: allow to set cwd

### LINKED ISSUE #178: Allow to set the CWD on the linter
Currently, nvim-lint sets CWD of the linter with the CWD of vim itself. But some linters (like mypy) work well only when launched from the project root directory.

So if `lcd` is used to go somewhere, mypy may not find the project configuration and/or some modules.

I have no idea what the perfect default is `vim.fn.getcwd()` or the project root directory, it really depends on the project and the tool I guess.

But I think the CWD of the linter should be configurable (I play using `lcd` just before launching the linter, but it's a bit hackish).

So if you also think it's a good idea to make CWD configurable, I will be happy to propose a Pull Request.

I imagined to allow something like:
```
require("lint.linters.mypy").cwd = "my path"
```

sileht:
Fixes #178

### DIFF
--- a/lua/lint/parser.lua
+++ b/lua/lint/parser.lua
@@ -46,7 +46,7 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
   severity_map = severity_map or {}
   -- Like vim.diagnostic.match but also checks if a `file` group matches the buffer path
   -- Some linters produce diagnostics for the full project and this should only produce buffer diagnostics
-  local match = function(buffer_path, line)
+  local match = function(linter_cwd, buffer_path, line)
     local matches = { line:match(pattern) }
     if not next(matches) then
       return nil
@@ -56,7 +56,12 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
       captures[groups[i]] = match
     end
     if captures.file then
-      local path = vim.fn.fnamemodify(captures.file, ':p')
+      local path
+      if vim.startswith(captures.file, '/') then
+        path = captures.file
+      else
+        path = vim.fn.simplify(linter_cwd .. '/' .. captures.file)
+      end
       if path ~= buffer_path then
         return nil
       end
@@ -84,14 +89,14 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
     end
     return vim.tbl_extend('keep', diagnostic, defaults or {})
   end
-  return function(output, bufnr)
+  return function(output, bufnr, linter_cwd)
     if not vim.api.nvim_buf_is_valid(bufnr) then
       return {}
     end
     local result = {}
     local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":p")
     for _, line in ipairs(vim.fn.split(output, '\n')) do
-      local diagnostic = match(buffer_path, line)
+      local diagnostic = match(linter_cwd, buffer_path, line)
       if diagnostic then
         table.insert(result, diagnostic)
       end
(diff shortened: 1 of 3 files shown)


mfussenegger:
Is this a use-case that you want to set the `cwd` *per* linter? I doubt there is a sensible `cwd` default other than `vim.fn.getcwd`, so users who want to change the working directory would then always have to modify all the linters they use?

How about adding an `opts` arguments to `try_lint` with a `cwd` property?
Not sure I like how impactful this is.

sileht:
> Is this a use-case that you want to set the `cwd` _per_ linter? 

I have no idea, I guess no. 

> I doubt there is a sensible `cwd` default other than `vim.fn.getcwd`, so users who want to change the working directory would then always have to modify all the linters they use?

In my previous vim experience, when I have used null-ls, ALE, (any older plugins I used)... the default was always the "project root directory" if found or buffer cwd if not. This has always worked out of the box for me.


> How about adding an `opts` arguments to `try_lint` with a `cwd` property? Not sure I like how impactful this is.

That should work for me.

sileht:
I updated the PR to pass the option to `try_lint()`

mfussenegger:
> In my previous vim experience, when I have used null-ls, ALE, (any older plugins I used)... the default was always the "project root directory" if found or buffer cwd if not. This has always worked out of the box for me.


neovim doesn't have the concept of a "project root directory".

There is actually work going on that might result in some kind of project module that would introduce such a concept. That's also the reason why I'm holding off on this for now.

sileht:
> > In my previous vim experience, when I have used null-ls, ALE, (any older plugins I used)... the default was always the "project root directory" if found or buffer cwd if not. This has always worked out of the box for me.
> 
> neovim doesn't have the concept of a "project root directory".
> 
> There is actually work going on that might result in some kind of project module that would introduce such a concept. That's also the reason why I'm holding off on this for now.

Exactly, that is why I just introduced the configuration of `cwd`, I don't think a tool like nvim-lint should implement any "project root directory" lookup. Personally, I used a lspconfig helper to do that.


To give you an idea of my current setup, I have:

```
autocmd BufWritePost <buffer> lua require("post_write_tools").run_linter()
```

And `run_linter()` does something like this:

```
function M.run_linter()
  local lspconfig = require("lspconfig")
  local rootdir = lspconfig.util.root_pattern(".git")(vim.api.nvim_buf_get_name(0))
  local env = {PATH = rootdir .. "/.tox/pep8/bin/:" .. rootdir .. "/node_modules/.bin/:" .. vim.fn.getenv("PATH"}
  require("lint.linters.flake8").env = env
  require("lint.linters.mypy").env = env
  require("lint.linters.eslint").env = env
  require("lint").try_lint(nil, {cwd = rootdir})
end
```

juanramirezc2:
I work with a monorepo structure using eslint and nested `.eslintrc` this will be pretty useful. 

right now eslint only works if I open vim from one of the subfolders. 
When i open vim from the root it doesn't pick the nested config 😞

juanramirezc2:
Vscode has a `working directories` option https://github.com/microsoft/vscode-eslint/issues/1170

mfussenegger:
@sileht would you be up for rebasing this? Sorry that I let this slide for so long. I wanted to wait out a discussion around some project system on neovim core (that it didn't happen), and then got sidetracked with plenty of other stuff.

I think in general adding the `opts` with the `cwd` property is good

sileht:
@mfussenegger done.

lua/lint/parser.lua
```diff
@@ -56,7 +56,13 @@ function M.from_pattern(pattern, groups, severity_map, defaults)
       captures[groups[i]] = match
     end
     if captures.file then
-      local path = vim.fn.fnamemodify(captures.file, ':p')
+      --local path = vim.fn.fnamemodify(captures.file, ':p')
```
mfussenegger, line 59:
```suggestion
```

mfussenegger:
Wow that was quick. Left a minor suggestion, otherwise this seems good to go.

sileht:
> Wow that was quick. Left a minor suggestion, otherwise this seems good to go.

I use it every day, so it was almost up to date

PR #181, opened 2022-02-09
fix(env): when PATH is passed via env

sileht:
if linter.env set PATH, env stay nil, then the copy of the env doesn't
work.


mfussenegger (commented):
This would always set `env` to an empty dict if the linter has no explicit `env` set - meaning that no environment variables at all would be set and the linter call would most likely fail because it cannot lookup the executable without at least `PATH`

Issue #186, opened 2022-02-14
Incorrect underlining (byte-based instead of glyph-based)


cpkio:
It seems underlining is byte-wise, not character-wise, which results in erroneous underlines.

### Vale JSON content
```
    {
      "Action": {
        "Name": "replace",
        "Params": null
      },
      "Check": "lexic.CheckBox",
      "Description": "",
      "Line": 336,
      "Link": "",
      "Message": "Использовать 'check box' вместо 'checkbox'",
      "Severity": "error",
      "Span": [
        7,
        14
      ],
      "Match": "checkbox"
    }
```

### Result
![2022-02-14_113207](https://user-images.githubusercontent.com/1917113/153828016-b9bdde00-0c4b-4f15-9ae9-1cf99f3a7233.png)

### Expected
Should underline the `checkbox` text.

cpkio:
This has been discussed here https://github.com/neovim/neovim/issues/17401

As far as I see, it will be `vale` to output byte-based offsets, or `nvim-lint` will have to recalculate diagnostics position: take the line form linter output, find the matched text (returned in JSON in `Match` field) per-byte, and push real byte-values to `vim.diagnostic`.

mfussenegger:
Merged https://github.com/mfussenegger/nvim-lint/pull/187

PR #187, opened 2022-02-14
Quick fix for incorrect highlighting with Vale linter

cpkio:
Made a quick fix for #186, hope it helps. No multi-line support, because I think Vale doesn't output multi-line errors.


mfussenegger (commented):
I think `string.find` will be unreliable if the same error occurs multiple times within a line.

This could be changed to convert the character to byte offset using something like `vim.str_byteindex` but I don't know what kind of notion of character vale uses. (That's the reason it just assumed byte offset to begin with - it doesn't document it)

cpkio:
@mfussenegger 
> what kind of notion of character

What do you mean? English is not my native language, sorry)

PR #193, opened 2022-03-08
Add linter for lacheck

santisoler:
Create a new linter that parses the outputs of lacheck, a linter for
LaTeX files. This linter only returns line number and a message through
stdout by passing the .tex file as input. lacheck is not capable of
getting the input through stdin.

### DIFF
--- a/lua/lint/linters/lacheck.lua
+++ b/lua/lint/linters/lacheck.lua
@@ -1,0 +1,15 @@
+-- path/to/file, line <linum>: <message>
+local pattern = '[^:]+, line (%d+):(.+)'
+local groups = { 'lnum', 'message' }
+
+return {
+  cmd = 'lacheck',
+  stdin = false,
+  args = {},
+  stream = stdout,
+  ignore_exitcode = false,
+  parser = require('lint.parser').from_pattern(pattern, groups, nil, {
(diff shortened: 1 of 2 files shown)


santisoler:
Thanks for creating this plugin! I've started using it today and it was really easy to add a new linter, even for someone with almost no knowledge of lua. Great job everyone!

lua/lint/linters/lacheck.lua
```diff
@@ -0,0 +1,15 @@
+-- path/to/file, line <linum>: <message>
+local pattern = '[^:]+, line (%d+):(.+)'
+local groups = { 'lnum', 'message' }
+
+return {
+  cmd = 'lacheck',
+  stdin = false,
+  args = {},
+  stream = stdout,
```
mfussenegger, line 9:
```suggestion
  stream = 'stdout',
```

mfussenegger (commented):
thanks for the PR. I think the `stream` definition is wrong as `stdout` isn't a variable but needs to be a string. Otherwise this looks good to me

lua/lint/linters/lacheck.lua
santisoler: Nice catch! Thanks @mfussenegger

PR #197, opened 2022-03-28
[WIP][Feature] Solhint

Dimfred:
currently work in progress, there are still some cases missing, just already opening it if someone wants to help or looks for it

- solhint parser (solidity)
- stdin could be supported, tho it seems to be bugging in combination with a json output see [here](https://github.com/protofire/solhint/issues/296)

Also what I noticed is that `DiagnosticUnderline*` highlights are not working, probably an issue on my side. Do you have any suggestions here?

### DIFF
--- a/lua/lint/linters/solhint.lua
+++ b/lua/lint/linters/solhint.lua
@@ -14,15 +14,24 @@ return {
   ignore_exitcode = true,
   parser = function(output, bufnr)
     local diagnostics = {}
-
     if output == "" then
       print("solhint: parse error")
       return diagnostics
     end
 
+    if output == "Failed to load a solhint's config file." then
+      print(output)
+      return diagnostics
+    end
+
     local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":~:.")
 
-    output = vim.json.decode(output)[1]
+    output = vim.json.decode(output)
+    if output == nil or table.getn(output) == 0 then
+      return diagnostics
+    end
+    output = output[1]
+
     for _, item in ipairs(output.reports or {}) do
       if not item.path or vim.fn.fnamemodify(item.path, ":~:.") == buffer_path then
         table.insert(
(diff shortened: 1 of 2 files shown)


lua/lint/linters/solhint.lua
```diff
@@ -0,0 +1,61 @@
+local vim = vim
```
mfussenegger, line 1:
```suggestion
```

This shouldn't be needed

lua/lint/linters/solhint.lua
```diff
@@ -0,0 +1,61 @@
...
+  parser = function(output, bufnr)
+    local diagnostics = {}
+    if output == "" then
+      print("solhint: parse error")
+      return diagnostics
+    end
+
+    if output == "Failed to load a solhint's config file." then
+      print(output)
+      return diagnostics
```
mfussenegger, line 24: Maybe use `vim.notify` instead of `print` or you could also add a diagnostic entry in the first line that shows the information.

lua/lint/linters/solhint.lua
```diff
@@ -0,0 +1,61 @@
...
+    end
+
+    local buffer_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":~:.")
```
mfussenegger, line 27: I think the `fnamemodify` call might cause problems if users set a custom `cwd` when calling `try_lint` or `lint`.

lua/lint/linters/solhint.lua
Dimfred: guess that was from a time where I had no vim lsp support will be removed.

mfussenegger:
Closing this for now as it appears stale.  Please reopen if you progressed on it

Dimfred:
Hey hey,
as always lost the time. Saw that the whole interface of `nvim-lint` changed, it is so much easier now. I just pushed my updates, looks like all the other linters I'd say. 
Should I open a new PR, or will you reopen?

PR #207, opened 2022-04-25
Call `on_done` callback in `on_exit` handler

### LINKED ISSUE #185: `on_done` not being called
When adding support for Elixir's [credo](https://hexdocs.pm/credo/overview.html), the `parser` function is not being called.

```lua
require('lint').linters.credo = {
  cmd = 'mix',
  stdin = true,
  args = { 'credo', 'suggest', '--format', 'json', '--read-from-stdin' },
  stream = 'stdout',
  ignore_exitcode = true, -- exits with severity code
  parser = function(output)
    error(vim.inspect(output)) -- never getting here
  end
  }
  ```
  
It looks like `parser.on_chunk` inside of the generated function from `read_output` is being called with the full content, but `parser.on_done` is never being called, so no diagnostics are being generated even though all of the output is provided.

In the luv does, I noticed that it [explicitly mentions](https://github.com/luvit/luv/blob/master/docs.md#uvread_startstream-callback) `EOF`.

> When we've reached EOF, data will be nil.

My assumption is that credo is not outputting an explicit EOF, so luv never calls the callback a final time.

I am not very familiar with luv, and if my assumption is correct, maybe there is a way to perform a final call inside of the `on_exit` callback of `uv.spawn`?

Thank you!

### PR DESCRIPTION
**Pull request by nifoc**
This works around an issue where the `on_done` callback was not always
called by the `read_output` function.

The most likely cause for this is that some linters do not generate an
EOF (that luv detects), therefore never triggering the `else` branch in
`read_output`.

The `on_done` is now called in two places:

  1. The existing `read_output` branch
  2. The `on_exit` handler

`on_done` has been extended to detect if it has already been called and
do nothing on the second invocation.

Closes #185

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -39,6 +39,16 @@ local namespaces = setmetatable({}, {
 })
 
 
+local function publish(bufnr, ns)
+  return function(diagnostics)
+    -- By the time the linter is finished the user might have deleted the buffer
+    if api.nvim_buf_is_valid(bufnr) then
+      vim.diagnostic.set(ns, bufnr, diagnostics)
+    end
+  end
+end
+
+
 local function read_output(bufnr, parser, publish_fn)
   return function(err, chunk)
     assert(not err, err)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by nifoc** [comment]
It looks like the test failure is caused by the `on_exit` handler not being called for the `compiler` linter.

Not sure about the exact cause yet, but it's the same on `master` (minus the test failure, because `on_done` is being called elsewhere).

**Comment by nifoc** [comment]
Failing test should be fixed. Handling `both` (`stdin` and `stdout`) caused problems, because the `on_done` callback that is "added" by the `split` helper was never being called.

I handled this by basically calling `on_done` twice (in some cases) and detecting if it has already been called. Not sure if that's the best solution, but it certainly was the easiest way to handle `both`.

**Comment by mfussenegger** [comment]
I wonder if the issue is caused by closing the pipes too early.

Could you try if something like this would also fix the problem?

```patch
diff --git a/lua/lint.lua b/lua/lint.lua
index 7a372c2..b929f69 100644
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -165,12 +165,16 @@ function M.lint(linter)
     detached = false
   }
   local cmd = eval_fn_or_id(linter.cmd)
   assert(cmd, 'Linter definition must have a `cmd` set: ' .. vim.inspect(linter))
   handle, pid_or_err = uv.spawn(cmd, opts, function(code)
-    stdout:close()
-    stderr:close()
+    stdout:shutdown(function()
+      stdout:close()
+    end)
+    stderr:shutdown(function()
+      stderr:close()
+    end)
     handle:close()
     if code ~= 0 and not linter.ignore_exitcode then
       print('Linter command', cmd, 'exited with code', code)
     end
   end)
```

nifoc:
That fixes the issue I had with `deadnix` and is so much nicer than my PR. Awesome!

Do you want me to open another PR or do you just want to commit this simple change? Both would obviously be totally fine by me.

mfussenegger:
> Do you want me to open another PR or do you just want to commit this simple change?

yes if you want to, you could create the PR. I don't mind either way.

Issue #203, opened 2022-04-21
golangci-lint not returning any results


keyboard-clacker:
Hey there! Not sure if this is the right place to put this bug, but I'm trying to use nvim-lint as my linter in go, and I'm having trouble setting args to pass to golangci-lint. Here's my config:

```lua
use({
	"mfussenegger/nvim-lint",
	config = function()
		require("lint").linters_by_ft = {
			go = { "golangcilint" },
		}

		local golangcilint = require("lint.linters.golangcilint")
		golangcilint.append_fname = true
		golangcilint.args = {
			"run",
			"--out-format",
			"json",
			"--config",
			"~/workspace/someDir/.golangci.yml",
			"--timeout",
			"5m",
		}
	end,
})
```

does anything look particularly off here? When I run the command that this _should_ run, aka: `golangci-lint run --out-format json --config ~/workspace/someDir/.golangci.yml --timeout 5m`, it properly outputs a bunch of errors, but I see no output or errors in nvim. I don't even see any errors or anything in my diagnostics when removing those args or any more specific config 

Thanks in advance! I really hope to be able to get this up and running and integrated with nvim native lsp

Cheers,
Kyle

mfussenegger:
Did you setup a autocmd to trigger `try_lint` or tried to invoke it manually?


This is what the readme recomments:

```vimL
au BufWritePost <buffer> lua require('lint').try_lint()
```

keyboard-clacker:
Yeah @mfussenegger

mfussenegger:
Maybe your configuration is not picked up. The `~` in `"~/workspace/someDir/.golangci.yml",` is not automatically expanded. You need to wrap it in a vim.fn.expand call or pass in the absolute path.

keyboard-clacker:
@mfussenegger is that relevant even though when calling golangci-lint binary directly, I can pass relatively paths like that? And my configuration is most definitely not getting picked up, that's the whole problem haha.

keyboard-clacker:
@mfussenegger I see, no I still don't see any document diagnostics or visual indication that anything works when expanding that path either. And when logging out the opts to the linter I see it's passing the expanded path

keyboard-clacker:
Also for what it's worth, I put a log in `nvim-lint/lua/lint/parser.lua`'s `accumulate_chunks` method, the `on_chunk` call, and it appears that that isn't even getting hit. I think something is borked about reading from the output or something with golangcilint

[...]

PR #208, opened 2022-04-27
Update golangcilint to read output from stdout and stderr

keyboard-clacker:
Fix for #203 

My understanding of the problem is that golangci-lint ouputs the EOF to stdout, and the actual errors to stderr, so this updates the stream to be both. I also changed it to append the filename since some go projects can be pretty large, but happy to take this part out since the user can also set this themselves if they'd like.

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -13,7 +13,7 @@ return {
     '--out-format',
     'json',
   },
-  stream = 'stdout',
+  stream = 'both',
   ignore_exitcode = true,
   parser = function(output, bufnr)
     if output == '' then


mfussenegger:
Not appending the filename was intentional, see https://github.com/mfussenegger/nvim-lint/commit/f632b0bc141ba6698a43425af43a2362eef9feba

> My understanding of the problem is that golangci-lint ouputs the EOF to stdout, and the actual errors to stderr, so this updates the stream to be both

That's not how pipes and EOF words. EOF is not data that is sent from an application, but a signal from the system that is afaik sent as soon as the pipe is closed & there are no more references to the file descriptors.

keyboard-clacker:
Oh interesting. I'll remove the fname setting then. And that's strange then, yeah it was just a guess for how it was working but it doesn't work with either stdout or stderr, so this seems to be the right solution regardless of reason

mfussenegger:
Can you try if your issue still occurs with https://github.com/mfussenegger/nvim-lint/pull/209 included and without the change in this PR?

keyboard-clacker:
Ah hahh, yes it works with that as well. Closing this issue since I guess it was closing pipes too early

PR #209, opened 2022-04-27
Prevent closing stdout and stdin pipes too early

### LINKED ISSUE #185: `on_done` not being called

nifoc:
Otherwise the `on_done` callback (which calls the `parser` function)
would sometimes not be called.

Closes #185

As discussed in #207


mfussenegger:
Thanks for pushing this and testing out this alternative solution :tada:

PR #212, opened 2022-04-30
feat: add glslc

theHamsta:
WIP: because this linter requires a output file. Also it can only guess the type of the shader when shader has typical file extension like `*.vert` `*.frag` or a pragma in the shader (https://github.com/google/shaderc/tree/main/glslc#311-shader-stage-specification)

### DIFF
--- a/lua/lint/linters/glslc.lua
+++ b/lua/lint/linters/glslc.lua
@@ -1,0 +1,18 @@
+-- path/to/file:line: severity: 'offending part of source code': message
+local pattern = "%s*([^:]+):(%d+): (.+): (.+: .+)"
+local groups = { "file", "lnum", "severity", "message" }
+
+return {
+  cmd = "glslc",
+  stdin = false,
+  args = {"-o", "-"}, -- "-" represents output of compilation result to stdout
+  ignore_exitcode = true,
+  stream = "stderr",
+  parser = require("lint.parser").from_pattern(pattern, groups, {
+    error = vim.diagnostic.severity.ERROR,
+    warning = vim.diagnostic.severity.WARN,
+  }, {
+      ["source"] = "glslc",
+      ["severity"] = vim.diagnostic.severity.WARN,
+    }),
+}
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ Other dedicated linters that are built-in are:
 | [Flake8][13]                 | `flake8`       |
 | [flawfinder][35]             | `flawfinder`   |
 | [Golangci-lint][16]          | `golangcilint` |
+| [glslc][glslc]               | `glslc`        |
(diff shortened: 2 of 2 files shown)


lua/lint/linters/glslc.lua
```diff
@@ -0,0 +1,30 @@
...
+    tmp_file = cache_dir .. "\\nvim-lint_glslc.spv"
+  else
+    error('Could not find folder of temporary file for glslc linter')
+  end
+end
+
+return {
+  cmd = "glslc",
+  stdin = false,
+  args = {"-o", tmp_file},
```
theHamsta, line 20: `tmp_file` could maybe be provided by the plugin if needed by more linters.

mfussenegger: How difficult would it be to get a change into this linter to not require the output file? 

This is the first linter I encounter that requires something like this. 

(Not a blocker for this PR, but I wouldn't extend the framework to support this more easily)

lua/lint/linters/glslc.lua
```diff
@@ -0,0 +1,30 @@
...
+local pattern = "%s*([^:]+):(%d+): (.+): (.+: .+)"
+local groups = { "file", "lnum", "severity", "message" }
+
+local tmp_file
+if vim.fn.fn.has "win32" == 1 then
+  tmp_file = "/dev/null"
+else
+  local cache_dir = vim.fn.stdpath("cache")
+  if vim.loop.fs_stat(cache_dir) then
+    tmp_file = cache_dir .. "\\nvim-lint_glslc.spv"
```
mfussenegger, line 11: Does this file grow or is it overwritten with each run?

Otherwise it may be necessary to generate it on each use of the linter and clean up after.

mfussenegger (commented):
Left a few comments, but I'd also be fine merging as is (assuming it works)

lua/lint/linters/glslc.lua
theHamsta: it is overwritten each run. SPRIV files are typically a few kB big. We could also just not enable this linter on Windows while we don't have a better solution.

`glslc` is actually a compiler (by Google distributed with the Vulkan SDK). There are also validators for `glsl` (e.g. glslValidator) but most not as convenient as this one. It might be possible to achieve that it does not require an output, but that use case is kind of strange for a compiler and the change would only be possible for future versions of the tool. I'd imagine that the requirement for an output file would exist for more tools that aren't linters in the traditional sense but rather compilers that would produce output normally

theHamsta: It works at the moment, but would love to test it a few days.

theHamsta:
Update: found in the documentation that `-` as ouput represents `stdout`

Issue #215, opened 2022-05-11
Error running codespell: EMFILE: too many open files


AckslD:
I sometimes get the following error:
```
...share/nvim/site/pack/packer/start/nvim-lint/lua/lint.lua:181: Error running codespell: EMFILE: too many open files
```
(also for eg `flake8`, `mypy` etc).

Does anyone know the cause of this? I don't have that many files open.

mfussenegger:
Sounds as if file/process handles are not closed.

I added https://github.com/mfussenegger/nvim-lint/commit/0407c340a77380e4122dc349efa10fc846c928b4 to ensure the error case is properly cleaned up, but I suspect you weren't running into that?

Did this start recently - more specifically after https://github.com/mfussenegger/nvim-lint/commit/d754a5f062ca62bc8c3e7ecee7d514ba1dfc4b23 was merged?

AckslD:
Thanks for your reply @mfussenegger!

It started relatively recently, could be since d75a5f but not certain.

I trigger
```lua
require('lint').try_lint()
require('lint').try_lint({"codespell"})
````
on `BufEnter` and `BufWritePost`. I do have a habit of saving quite a lot but not sure that's the issue.

AckslD:
Hi @mfussenegger, I haven't had time to try to revert the commit yet, but let you know once I do and if it helps.

My neovim version is:
NVIM v0.8.0-dev+227-g6613f58ceb

and luv version:
76544

dbarrosop:
I am also having this problem with the golangci-lint linter, I just updated neovim so I will wait and see if it happens again and if it does I will revert https://github.com/mfussenegger/nvim-lint/commit/d754a5f062ca62bc8c3e7ecee7d514ba1dfc4b23 and test.

AckslD:
I'm starting to suspect that it's a combination of plugins starting jobs/threads. I got the same error with one of my own plugins which uses `vim.fn.jobstart`. Does any of you know if there is a way to check how many or what jobs/threads are opened/started with `vim.fn.jobstart` and/or `vim.loop`?

mfussenegger:
By checking the contents of `/proc/<nvim-pid>/fd` it is visible that nvim-lint leaks descriptors with the change made in https://github.com/mfussenegger/nvim-lint/commit/d754a5f062ca62bc8c3e7ecee7d514ba1dfc4b23

[...]

PR #218, opened 2022-05-28
Ensure stdout/stderr pipes get closed

mfussenegger:
Should fix https://github.com/mfussenegger/nvim-lint/issues/215

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -57,6 +57,10 @@ local function start_read(stream, stdout, stderr, bufnr, parser, ns)
     if api.nvim_buf_is_valid(bufnr) then
       vim.diagnostic.set(ns, bufnr, diagnostics)
     end
+    stdout:shutdown()
+    stdout:close()
+    stderr:shutdown()
+    stderr:close()
   end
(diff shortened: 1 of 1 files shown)


mfussenegger:
@nifoc Could you try if this still works with `deadnix` ? The change in https://github.com/mfussenegger/nvim-lint/pull/209 unfortunately causes file descriptors to leak

nifoc:
This seems to break `deadnix` again, but I suppose leaking file descriptors is worse than one linter (that isn't even part of the default configuration) breaking.

mfussenegger:
> This seems to break `deadnix` again, but I suppose leaking file descriptors is worse than one linter (that isn't even part of the default configuration) breaking.

Thank you for checking. I pushed another commit that still fixes the leak and might keep `deadnix` working. Could you give it another try?

nifoc:
Just tested with [f510830](https://github.com/mfussenegger/nvim-lint/pull/218/commits/f51083092875dad4da718313e0385db22596126d) and `deadnix` still works as expected 🥳

PR #223, opened 2022-06-15
feat: add DirectX Shader Compiler

theHamsta:
Seams to work. Will try out for a week or so...

Diff in README is huge because I needed to shift the separator by one space because "DirectX Shader Compiler" was too long.


mfussenegger (commented):
Looks good to me. Please ping me once it's ready to merge.

theHamsta:
I think this should be ready-to-merge!

PR #226, opened 2022-06-22
Fix #225: hlint fails with non-zero exit-code unless there are no hints

### LINKED ISSUE #225: hlint fails with non-zero exit-code unless there are no hints
When calling `lua require('lint').try_lint()` with hlint, it fails because hlint returns a non-zero exit-code unless there are no hints.

MrcJkb:
Fixes #225


mfussenegger:
Thanks :+1:

PR #232, opened 2022-07-22
Added `lint-openapi`

cpkio:
Added support for [IBM OpenAPI validator](https://github.com/IBM/openapi-validator).

By default `lint-openapi` output separate lists of error and warning messages, so theres' two `for` loops (for each list). Only Errors and Warnings will be displayed for now.

`lint-openapi` cannot handle `\` or `\\` in file paths, so I had to add a function to replace `\` to `/`.


mfussenegger:
Thanks, looks good to me

PR #239, opened 2022-08-12
Fix column for pylint

seblj:
I noticed it didn't have any underline/undercurl etc for pylint. This should fix it.

First I tried to just add the end_column, but noticed the starting column was one off.
Pasted Graphic 1

## Result
Pasted Graphic 2


seblj:
I'll update the tests as soon as possible

PR #255, opened 2022-10-01
feat: add support for jsonlint

### LINKED ISSUE #246: Consider supporting jsonlint
Being able to lint JSON would be helpful. 

Here is the relevant config from null-ls: https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/jsonlint.lua

prichrd:
fixes #246


mfussenegger:
Looks good,  thanks

PR #263, opened 2022-10-23
[Feature] Add linter option, append_dname

lwyw:
Append folder name to `golangci-lint` to lint current package only instead of the entire workspace.

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -8,6 +8,7 @@ local severities = {
 return {
   cmd = 'golangci-lint',
   append_fname = false,
+  append_dname = true,
   args = {
     'run',
     '--out-format',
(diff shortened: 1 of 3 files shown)


lua/lint/linters/golangcilint.lua
```diff
@@ -8,6 +8,7 @@ local severities = {
 return {
   cmd = 'golangci-lint',
   append_fname = false,
+  append_dname = true,
```
mfussenegger, line 11: Could you instead change the args to:

```lua
  args = {
    'run',
    '--out-format',
    'json',
    function()
      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":h")
    end
  },
```

And remove the `append_dname` option?
I'd like to refrain from adding more general options until it is required by more than just one linter.

lwyw: https://github.com/mfussenegger/nvim-lint/pull/265/files

lwyw:
close in favor of https://github.com/mfussenegger/nvim-lint/pull/265

PR #266, opened 2022-10-31
docs: update README to fix typo

nat-418:
In my original PR I accidentally put `Tcl` instead of `Nagelfar` for the tool name. This commit corrects that error.


mfussenegger:
Thanks

PR #268, opened 2022-11-04
[Feature] Add Bitbake/oelint-adv support

antznin:
This PR adds support for the [Bitbake](https://en.wikipedia.org/wiki/BitBake) programming language using the oelint-adv syntax analysis tool.


mfussenegger:
Thanks

PR #270, opened 2022-11-18
Add psalm as linter

danrot:
I copied this part of my nvim configuration. Is that everything which is needed? Or do I have to additionally register it somehow?

### DIFF
--- a/lua/lint/linters/psalm.lua
+++ b/lua/lint/linters/psalm.lua
@@ -1,29 +1,29 @@
 return {
-	cmd = 'psalm',
-	args = {
-		'--output-format=json',
-		'--no-progress',
-	},
-	parser = function(output)
-		if output == nil then
-			return {}
-		end
+  cmd = 'psalm',
+  args = {
+    '--output-format=json',
+    '--no-progress',
+  },
+  parser = function(output)
+    if output == nil then
+      return {}
+    end
 
-		local messages = vim.json.decode(output)
-		local diagnostics = {}
+    local messages = vim.json.decode(output)
+    local diagnostics = {}
 
-		for _, message in ipairs(messages or {}) do
-			table.insert(diagnostics, {
-				lnum = message.line_from - 1,
-				end_lnum = message.line_to - 1,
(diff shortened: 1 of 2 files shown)


mfussenegger:
Thanks for the PR.
Adding an entry to the table in the readme would be nice, otherwise this is all that's needed.

The linter is unhappy about indentation, other than that this looks good to merge to me.

danrot:
All done 🙂

lua/lint/linters/psalm.lua
```diff
@@ -0,0 +1,29 @@
...
+
+    for _, message in ipairs(messages or {}) do
+      table.insert(diagnostics, {
+        lnum = message.line_from - 1,
+        end_lnum = message.line_to - 1,
+        col = message.column_from - 1,
+        end_col = message.column_to - 1,
+        message = message.message,
+        source = 'psalm',
+        severity = message.severity,
```
mfussenegger, line 23: Sorry I missed this initially, `severity` must be one of `:h diagnostic-severity`. Unless `message.severity` happens to be a number from 1-4 that happens to align 1:1 with the vim.diagnostic definition it needs to be mapped.

Other linters usually have some kind of mapping:

```
local severities = {
  error = vim.diagnostic.severity.ERROR,
  warning = vim.diagnostic.severity.WARN,
  information = vim.diagnostic.severity.INFO,
  hint = vim.diagnostic.severity.HINT,
  suggestion = vim.diagnostic.severity.HINT,
}
```

And then use something like `severities[message.severity] or vim.diagnostic.severity.WARN` to retrieve it and have a default value.

danrot: I have found info and error diagnostics, and the both seem to work as I have implemented it currently:

image

mfussenegger: Alright,  if it turns out to miss some states or map it incorrectly we can still extend it later.  Merged the PR as is

danrot: Thank you!

PR #276, opened 2022-11-25
Add cfn_nag

jpfender:


lua/lint/linters/cfn_nag.lua
```diff
@@ -0,0 +1,34 @@
+local severity_map = {
+  ["WARN"] = vim.diagnostic.severity.WARN,
+  ["FAIL"] = vim.diagnostic.severity.ERROR,
+}
+
+return {
+  cmd = "cfn_nag_scan",
+  stdin = false,
+  args = { "2>/dev/null", "--output-format", "json", "--input-path" },
```
mfussenegger, line 9:
```suggestion
  args = {"--output-format", "json", "--input-path" },
```

nvim-lint launches the commands directly without intermediate shell, `2>/dev/null` would be passed as literal value to the linter.

jpfender: Thanks, I wasn't aware of that! I've removed the argument.

PR #279, opened 2022-12-13
Add code field to Pylint's diagnostics

rafikdraoui:
The value can be used when rendering diagnostic information. For example, it is used by default in the diagnostic float window since https://github.com/neovim/neovim/pull/21130.

c.f. `:help diagnostic-structure`

### DIFF
--- a/lua/lint/linters/pylint.lua
+++ b/lua/lint/linters/pylint.lua
@@ -30,7 +30,7 @@ return {
           end_col = end_column,
           severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
           message = item.message,
-          code = item.symbol,
+          code = item['message-id'],
           user_data = {
             lsp = {
               code = item['message-id'],
--- a/lua/lint/linters/pylint.lua
+++ b/lua/lint/linters/pylint.lua
@@ -30,6 +30,7 @@ return {
           end_col = end_column,
           severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
           message = item.message,
+          code = item.symbol,
           user_data = {
             lsp = {
               code = item['message-id'],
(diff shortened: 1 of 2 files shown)


lua/lint/linters/pylint.lua
```diff
@@ -30,6 +30,7 @@ return {
           end_col = end_column,
           severity = assert(severities[item.type], 'missing mapping for severity ' .. item.type),
           message = item.message,
+          code = item.symbol,
```
mfussenegger, line 33: Shouldn't this be the `message-id` field instead of symbol?

rafikdraoui: The `symbol` field is the human-readable message (for example `undefined-variable`), while `message-id` is the code (for example `E0602`). I believe the former is more useful, but I don't know if that's a widespread preference.

For example, one use case for knowing a diagnostic code is to add a one-off "disable linting" pragma for it. In pylint, this ends up looking as a comment of the form `pylint: disable=x`, where `x` can either be the symbolic name or the ID of the linting rule. In that situation, I believe that using the human-readable name would be better, since a reader can more easily understand what the disabled rule is about, without having to browse through Pylint's documentation to find it.

I'm aware that `user_data.lsp.code` is using the message ID, so maybe it makes sense to also use it for the `code` field for consistency.

mfussenegger: I'd go with the message-id code,  that's commonly used for the `code` property and I think it would make it consistent with other Python linters like flake8 or ruff.

The point about the human-readable version is a good one.  Maybe that would even justify a separate field in the diagnostics structure.  But that would've to be discussed in neovim first

mfussenegger:
Thanks

PR #283, opened 2023-01-05
Add option to use cppcheck with C

### LINKED ISSUE #282: Add option to use cppcheck with C rather than C++
Despite its name, cppcheck also works on C files. Please expose an option to use it with C. I'd do it myself but I can't find where specific linters get activated per filetype.

EleanorNB:
Nvm, it's in the config. I have a brain, I swear.

Closes #282

### DIFF
--- a/lua/lint/linters/cppcheck_c.lua
+++ b/lua/lint/linters/cppcheck_c.lua
@@ -1,0 +1,25 @@
+-- cppcheck <= 1.84 doesn't support {column} so the start_col group is ambiguous
+local pattern = [[([^:]*):(%d*):([^:]*): %[([^%]\]*)%] ([^:]*): (.*)]]
+local groups = { 'file', 'lnum', 'col', 'code', 'severity', 'message' }
+local severity_map = {
+  ['error'] = vim.diagnostic.severity.ERROR,
+  ['warning'] = vim.diagnostic.severity.WARN,
+  ['performance'] = vim.diagnostic.severity.WARN,
+  ['style'] = vim.diagnostic.severity.INFO,
+  ['information'] = vim.diagnostic.severity.INFO,
+}
+
+return {
+  cmd = 'cppcheck',
+  stdin = false,
+  args = {
+    '--enable=warning,style,performance,information',
+    '--language=c',
+    '--inline-suppr',
+    '--quiet',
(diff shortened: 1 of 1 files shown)


EleanorNB:
No I don't, this doesn't work.

EleanorNB:
Yes it does, actually. Works great.

mfussenegger (commented):
Maybe the existing linter definition could instead be extended to support both with something like this:

```diff
diff --git a/lua/lint/linters/cppcheck.lua b/lua/lint/linters/cppcheck.lua
index dd2a4a5..7bd399a 100644
--- a/lua/lint/linters/cppcheck.lua
+++ b/lua/lint/linters/cppcheck.lua
@@ -12,11 +12,17 @@ local severity_map = {
 return {
   cmd = 'cppcheck',
   stdin = false,
   args = {
     '--enable=warning,style,performance,information',
-    '--language=c++',
+    function()
+      if vim.bo.filetype == 'cpp' then
+        return '--language=c++'
+      else
+        return '--language=c'
+      end
+    end,
     '--inline-suppr',
     '--quiet',
     '--cppcheck-build-dir=build',
     '--template={file}:{line}:{column}: [{id}] {severity}: {message}',
   },
```

EleanorNB:
You're right, that is better. I'll get that pushed momentarily.

EleanorNB:
@mfussenegger Pushed and ready.

lua/lint/linters/cppcheck.lua
```diff
@@ -1,25 +1,31 @@
...
-    '--cppcheck-build-dir=build',
-    '--template={file}:{line}:{column}: [{id}] {severity}: {message}',
+    "--enable=warning,style,performance,information",
+    (function()
+      if vim.bo.filetype == "cpp" then
+        return "--language=c++"
+      else
+        return "--language=c"
+      end
+    end)(),
```
mfussenegger, line 23: This should stay a function to ensure it's lazy evaluated on use and not on import. Otherwise if you open first a cpp project and then a c project it would be frozen to c++.

EleanorNB: Right, sorry, didn't know that. Fixed.

mfussenegger:
Thanks, merged

PR #290, opened 2023-01-21
Add phpstan as linter

danrot:
I would like to contribute another linter for PHP. This one does only know about errors and only tell one line, so there is not a lot of information that can be returned. It also does not know about the concept of severity...

### DIFF
--- a/lua/lint/linters/phpstan.lua
+++ b/lua/lint/linters/phpstan.lua
@@ -1,0 +1,33 @@
+return {
+  cmd = 'phpstan',
+  args = {
+    'analyze',
+    '--error-format=json',
+    '--no-progress',
+  },
+  ignore_exitcode = true,
+  parser = function(output)
+    if output == nil then
+      return {}
+    end
+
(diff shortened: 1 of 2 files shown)


lua/lint/linters/phpstan.lua
```diff
@@ -0,0 +1,33 @@
...
+    '--error-format=json',
+    '--no-progress',
+  },
+  ignore_exitcode = true,
+  parser = function(output)
+    if output == nil then
+      return {}
+    end
+
+    local file = vim.json.decode(output).files[vim.api.nvim_buf_get_name(0)]
```
mfussenegger, line 14:
```suggestion
  parser = function(output, bufnr)
    if output == nil then
      return {}
    end
    local file = vim.json.decode(output).files[vim.api.nvim_buf_get_name(bufnr)]
```

The linting process runs asynchronous and by the time the parser function is called a user could've switched to another buffer.

danrot: Didn't know about the `bufnr` argument, but makes sense. Already updated.

mfussenegger:
Thanks

PR #291, opened 2023-01-27
Support for rust_analyzer

benmanbenman:
Simple rust_analyzer support. I don't know a lot about how rust_analyzer works so it can definitely be improved.

### DIFF
--- a/lua/lint/linters/rust_analyzer
+++ b/lua/lint/linters/rust_analyzer
@@ -1,0 +1,7 @@
+return {
+  cmd = 'rust_analyzer',
+  stdin = true,
+  args = {},
+  ignore_exitcode = true,
+  parser = require('lint.parser').from_errorformat('[%t] %. %f:%l:%c %m, [%t] %. %f:%l %m')
+}


mfussenegger:
Thanks for the PR. Could you also add the linter to the readme?

benmanbenman:
Just added a PR to update the readme

mfussenegger:
Would be good to have that in the same PR. It also looks like [rust-analyzer](https://rust-analyzer.github.io/) is a language server, not a linter? Or what is the `rust_analyzer` command referring to?

benmanbenman:
Hi, as far as I'm aware it's a language server with linting capabilities so if you don't want to include that's fair enough. Let me know if you're ok with including it and I'll try and combine them

mfussenegger:
As far as I can tell you can't use rust-analyzer directly as linter. The rust-analyzer process will expect LSP communication on stdio, and nvim-lint expects some sort of linting result on stdout or stderr.

The linting capabilities it provides have to be consumed via Neovim's LSP client

benmanbenman:
Fair enough. It works for me just as well as the built-in options and I've been using it for a while. I might try to implement one with cargo-clippy

PR #292, opened 2023-01-28
Update readme to include rust_analyzer

benmanbenman:


mfussenegger:
See https://github.com/mfussenegger/nvim-lint/pull/291#issuecomment-1407638101

PR #294, opened 2023-02-10
sqlfluff: Do not use stdin and ignore useless output

jpfender:
# Overview

The current version of the sqlfluff parser has two major issues:

- It calls sqlfluff using stdin, which causes issues when used in the context of dbt
- It errors out if it encounters non-JSON output, which (at least on my setup) will always be the case as sqlfluff produces some boilerplate output that cannot be suppressed.

# Details

When I call sqlfluff in my project the way it's called by the current parser (using stdin), I get the following output:

```
❯ cat path/to/file.sql | sqlfluff lint --format=json -
==== readout ====

WARNING    File /full/path/to/cwd/stdin was not found in dbt project
==== summary ====
violations:        0 status:         PASS
[]

```

When I pass the file as an argument instead, it works:

```
❯ sqlfluff lint --format=json path/to/file.sql
==== readout ====

==== summary ====
violations:        2 status:         FAIL
[{"filepath": "path/to/file.sql", "violations": [{"line_no": 14, "line_pos": 43, "code": "L029", "description": "Keywords should not be used as identifiers."}, {"line_no": 21, "line_pos": 41, "code": "L029", "description": "Keywords should not be used as identifiers."}]}]
```

This output is parseable, but it still comes with leading boilerplate, which the current parser can't deal with. It simply takes the whole stdout output as an error message and appends it to line 0:

Screenshot 2023-02-10 at 09 34 42

Therefore, we need to iterate through the output line by line (the actual JSON output will always be on a single line) and ignore everything that is not JSON. This works:

Screenshot 2023-02-10 at 09 39 40

# Summary

- We pass the file to sqlfluff directly instead of using stdin.
- We step through the output line by line and ignore everything that's not JSON.


mfussenegger (commented):
What version of sqlfluff are you using?


With sqlfluff 1.4.5 using stdin seems to work fine for me and I also don't get any `==== readout ====` markers.


```
echo "SELECT 1," | sqlfluff lint --format=json --dialect=postgres -
[{"filepath": "stdin", "violations": [{"line_no": 1, "line_pos": 9, "code": "L038", "description": "Trailing comma in select statement forbidden"}]}]
```


```
cat foo.sql | sqlfluff lint --format=json --dialect=postgres -
[{"filepath": "stdin", "violations": [{"line_no": 4, "line_pos": 1, "code": "L036", "description": "Select targets should be on a new line unless there is only one select target."}, {"line_no": 4, "line_pos": 1, "code": "L044", "description": "Query produces an unknown number of result columns."}, {"line_no": 5, "line_pos": 3, "code": "L003", "description": "Expected 1 indentation, found less than 1 [compared to line 04]"}, {"line_no": 7, "line_pos": 3, "code": "L003", "description": "Expected 1 indentation, found less than 1 [compared to line 06]"}]}]
```


> when used in the context of dbt

Can you elaborate on what that means?

jpfender:
Hi there, sorry for the two month wait, I got bogged down with other things.

It turns out that the readout markers were because of verbosity settings in our project, not because of any sqlfluff version differences.

However, reading from stdin still does not work with my setup. I am using sqlfluff version 2.0.2. I've reduced the changes to just not using stdin.

Issue #302, opened 2023-03-15
feature request: cancel running linters


FelipeLema:
hey, there... great plugin you've got here

I want to be able to kill processes for running linter. I've got some linters running on `TextChanged` and I'd like the last run to kill all previously running (since I'm only interested in the last diagnostics).

Would the maintainers be willing to accept a PR for this?

mfussenegger:
I think an option could be to return the uv handle from `lint` and then have `try_lint` either return a list of handles or maybe a table with a `cancel` function on it? Slight tendency for the latter as it provides more flexibility for future extensions

PR would be welcome

FelipeLema:
we can break it into two pieces:

- have `lint` and `try_lint` return handles for every `spawn`-ed process
- handle `buffer → version_number → handles` mapping, cancelling all handles of previous versions whenever a new version arrives (maybe optional defaulting to "cancel previous"?? like `try_lint({linters={…}, cancel_old=true})`)

The second one may be shooting itself in the foot by trying to, in some cases, handle a global linter in a per-buffer context. If it's too complicated, we can just leave part 1.

FelipeLema:
part 2/2 here: https://github.com/mfussenegger/nvim-lint/pull/307

PR #306, opened 2023-03-22
bubble up process handle to `try_lint()`

FelipeLema:
part 1 of #302


mfussenegger:
Thanks.

I changed the result from `try_lint` to wrap the handles in an additional table. That allows to extend the result. (E.g. could add a `cancel` function as well.

PR #307, opened 2023-03-23
cancel previously spawned processes ( fix #302 )

### LINKED ISSUE #302: feature request: cancel running linters

FelipeLema:
store process handles on each `try_lint()` call on a per-buffer basis, cancel processes that are still running on every `try_lint()` call

part 2/2 of #302

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -114,18 +114,14 @@ function M._resolve_linter_by_ft(ft)
   return vim.tbl_keys(dedup_linters)
 end
 
---- map from buffer to handles from processes spawned by previous try_lint() call
----@type table<number, uv_handle_t>
-local buffer_to_running_handles = {}
 
---- autocommand for cleaning up handles when a buffer is deleted
---- each buffer gets its own autocommand
----@type table<number, number>
-local buffer_handles_cleanup_au = {}
+--- Running processes by buffer -> by linter name
+---@type table<integer, table<string, uv_process_t>> bufnr: {linter: handle}
+local running_procs_by_buf = {}
+
 
 ---@param names? string|string[] name of the linter
 ---@param opts? {cwd?: string, ignore_errors?: boolean} options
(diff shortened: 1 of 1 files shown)


lua/lint.lua
```diff
@@ -150,9 +163,9 @@ function M.try_lint(names, opts)
       notify(handle_or_error --[[@as string]], vim.log.levels.WARN)
     end
   end
-  return {
-    handles = handles
-  }
+
+  buffer_to_running_handles[bufnr] = handles
+
```
mfussenegger, line 168: This would be leaking the handles as they never get cleaned up if a buffer is closed/deleted.

lua/lint.lua
```diff
@@ -140,6 +143,16 @@ function M.try_lint(names, opts)
...
     return linter
   end
+
+  -- kill any running process for this buffer
+  local bufnr = api.nvim_get_current_buf()
+  local previous_handles = buffer_to_running_handles[bufnr] or {}
+  for _, handle in ipairs(previous_handles) do
+    if handle and not handle:is_closing() then
+      handle:close()
+    end
```
mfussenegger, line 153: I'm not sure if calling `:close` is the right thing to do here as this can cause defunct processes. See https://github.com/libuv/libuv/issues/1911 

This would probably need to send a term signal instead.

FelipeLema:
comments have been adressed... handling memory leak would be too cumbersome to implement for v0.6.2

mfussenegger:
Pushed a change that removes the autocmds. Could you check if it still works?

(And I suspect it didn't really work before, afaik `handle:kill("SIGTERM")` is not sending the right signal, it must be lower case))

PR #309, opened 2023-04-03
Find stylelint in node_modules binaries first

cpof-tea:
Same as for ESLint: first try to find stylelint in node_modules binaries, fallback to system stylelint


mfussenegger:
Thanks

PR #311, opened 2023-04-05
Add fname as an argument to stylelint

cpof-tea:
This way it respects stylelint's file based overrides [^1].

[^1]: https://stylelint.io/user-guide/configure/#overrides

### DIFF
--- a/lua/lint/linters/stylelint.lua
+++ b/lua/lint/linters/stylelint.lua
@@ -18,7 +18,9 @@ return {
     "json",
     "--stdin",
     "--stdin-filename",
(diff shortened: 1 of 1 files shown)


mfussenegger:
I'm not sure why it's used in stylelint or in standarrb. For `vim.fn.expand` it resolves to the full path of the current buffer.

You could try something like this for example:

```
  args = {
    "-f",
    "json",
    "--stdin",
    "--stdin-filename",
    function()
      return vim.fn.expand("%:p")
    end,
  },
```

PR #312, opened 2023-04-07
Add option to skip executing missing linters

pitbuster:


mfussenegger:
Thanks for the PR but `try_lint` already takes a `ignore_errors` option.

PR #314, opened 2023-04-10
Fix: linters installed by npm doesn't work on Windows

igrep:
Problem
====

On Windows, some linters are installed with the shim command suffixed with ".cmd" to run by its interpreter programs (such as Node.js). In such a case, we have to call the shim command to correctly run on Windows, otherwise we get an error like below:

```
Error running eslint_d: ENOENT: no such file or directory
```

Solution
====

Make a wraper around for such linters to call the shim command using cmd.exe.

Note
====

I applied the wrapper function `call_with_cmd_exe_on_windows` to some linters installed by npm in haste. But I guess there is other linters that should be wrapped.

Related change
====

<https://github.com/neovim/nvim-lspconfig/pull/1522>

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -6,7 +6,7 @@ local severity_map = {
   ['warning'] = vim.diagnostic.severity.WARN,
 }
 
-return require('lint.util').call_with_cmd_exe_on_windows {
+return require('lint.util').inject_cmd_exe({
   cmd = function()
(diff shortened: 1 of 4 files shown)


igrep:
Sorry, I found this change is not working yet. Marked as draft.

leblocks:
I think this is a necessary fix, I myself using neovim on windows so this problem is known to me.

mfussenegger:
Thanks

PR #315, opened 2023-04-16
psalm: Prefer `./vendor/bin/psalm`

hex-cat-man:
Prefer a local installation as some other linter configs already do.
Psalm can also be installed as a phar, so I added a check for this as well.


mfussenegger:
Thanks

PR #317, opened 2023-04-18
hlint: add suggested change to message

mrcjkb:
The current implementation only shows the summary for hlint.

This adds the suggested change to the diagnostic message.

Before:

```haskell
concat $ map escapeC s -- Use concatMap
```

After:

```haskell
concat $ map escapeC s -- Use concatMap: concatMap escapeC s
```


mfussenegger:
Thanks :+1:

PR #318, opened 2023-04-22
Add Linux kernel's checkpatch.pl support

antznin:
As described in checkpatch.lua, the user either needs to adjust cmd from its config to provide a full path to checkpatch.pl, or checkpatch.pl should be available in the PATH.


mfussenegger:
Thanks

PR #323, opened 2023-07-13
Add Support for PHP's built-in linter

Geraint:
This one picks up basic syntax errors using PHP's built-in linter (`php -l`).

It would be good to add this to nvim-lint, since other tools support this:

- [ALE](https://github.com/dense-analysis/ale/blob/master/ale_linters/php/php.vim)
- [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/diagnostics/php.lua)


mfussenegger:
Thanks again

PR #324, opened 2023-07-16
Configure eslint_d to read from stdin

gualcasas:
eslint_d can read from stdin, but it's currently not configured to do so.

With this change, we can now lint on "TextChanged" and "InsertLeave"

### DIFF
--- a/lua/lint/linters/eslint_d.lua
+++ b/lua/lint/linters/eslint_d.lua
@@ -18,7 +18,7 @@ return require('lint.util').inject_cmd_exe({
   args = {
     '--stdin',
     '--stdin-filename',
-    vim.api.nvim_buf_get_name(0),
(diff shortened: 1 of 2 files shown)


lua/lint/linters/eslint_d.lua
```diff
@@ -15,8 +15,12 @@ return require('lint.util').inject_cmd_exe({
     end
     return 'eslint_d'
   end,
-  args = {},
-  stdin = false,
+  args = {
+    '--stdin',
+    '--stdin-filename',
+    vim.api.nvim_buf_get_name(0),
```
mfussenegger, line 21:
```suggestion
    function() return vim.api.nvim_buf_get_name(0) end,
```

Otherwise it would freeze the stdin-filename to the name of the buffer it had when `require` is used the first time.

PR #328, opened 2023-07-23
fix: cspell not work

hewenjin:
cspell lint have some issue, it have no diagnostics, and the parse function return wrong `end_col`  value. this PR fix these.

### DIFF
--- a/lua/lint/linters/cspell.lua
+++ b/lua/lint/linters/cspell.lua
@@ -1,18 +1,38 @@
+local efm = '%f:%l:%c - %m'
 return {
   cmd = 'cspell',
-  stdin = true,
   ignore_exitcode = true,
   args = {
     'lint',
     '--no-color',
     '--no-progress',
     '--no-summary',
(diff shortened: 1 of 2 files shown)


hewenjin:
I use cspell to check every buffer spell.j A suitable config is in here:
https://github.com/hewenjin/config/blob/6c3a91c2c1071689025ff170803d70290954734b/nvim/lua/plugin/lspconfig.lua#L183C1-L197

mfussenegger:
@cyntheticfox could you check the changes?

cyntheticfox:
On its face, it looks like the changes should work, but I don't have the time to make a config to test it at the moment. Only thing might also be removing the ignoring of the exit code, as it seems non-useless

hewenjin:
by removing `ignore_exitcode`, I got an error
```
Linter command `cspell` exited with code: 1
```
seems like cspell return exitcode 1 when it detected issue. so it shouldn't remove I guess

Issue #236, opened 2022-08-05
Feature Request: Support for HTML templetes linter


zefr0x:
There is a supported HTML linter, but now a days most people use templets to generate HTML, rather then just writing a normal HTML file.

There are one or more HTML templete generator for every server-side language as listed [here](https://en.wikipedia.org/wiki/Template_engine_(web)#Server-side_systems). Personaly I've used [jinja2](https://en.wikipedia.org/wiki/Jinja_(template_engine)) before in python.

> **Note** HTML Tidy (the current supported HTML linter) don't messes up Jinja templates after [this issue](https://github.com/htacg/tidy-html5/issues/310). So it's usable, and it will not show errors, it will just skip Jinja elements without trying to lint them.

**Here are some HTML templete generator  linters I found:** _(They support more then one templete generator)_
- [djLint](https://djlint.com/) (Active project and support more telemplets)
- [djhtml](https://github.com/rtts/djhtml)
- [curlylint](https://www.curlylint.org/)

mfussenegger:
Added curlylint with https://github.com/mfussenegger/nvim-lint/pull/289

PR #329, opened 2023-07-24
Added djlint support

Penaz91:
Greetings,

I have kept an eye on this project since I got to know that null-ls is going to be archived. I decided to add support for DjLint (although I couldn't test it, because it seems my environment doesn't currently allow me to use `try_lint` and have the results published in the diagnostic list).

This hopefully helps with #236 

Thank you for your attention!

### DIFF
--- a/lua/lint/linters/djlint.lua
+++ b/lua/lint/linters/djlint.lua
@@ -1,0 +1,19 @@
+local pattern = [[([^:]+):(%d+):(%d+):(%d+): (.*)]]
+local groups = { 'file', 'lnum', 'col', 'code', 'message' }
+
+local defaults = {
+  ['source'] = 'djlint',
+  ['severity'] = vim.diagnostic.severity.INFO
+}
+
+return {
+  cmd = 'djlint',
+  stdin = false,
+  args = {
+    '--linter-output-format',
+    '{filename}:{line}:{code}: {message}',
+  },
+  stream = 'both',
+  ignore_exitcode = true,
+  parser = require('lint.parser').from_pattern(pattern, groups, nil, defaults, {}),
+}
--- a/README.md
+++ b/README.md
@@ -89,6 +89,7 @@ Other dedicated linters that are built-in are:
 | [credo][credo]                     | `credo`           |
 | [cspell][36]                       | `cspell`          |
 | [curlylint][curlylint]             | `curlylint`       |
+| [djlint][djlint]                   | `djlint`          |
 | [erb-lint][erb-lint]               | `erb_lint`        |
 | [ESLint][25]                       | `eslint`          |
 | [eslint_d][37]                     | `eslint_d`        |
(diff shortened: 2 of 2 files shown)


lua/lint/linters/djlint.lua
```diff
@@ -0,0 +1,19 @@
...
+  ['source'] = 'djlint',
+  ['severity'] = vim.diagnostic.severity.INFO
+}
+
+return {
+  cmd = 'djlint',
+  stdin = false,
+  args = {
+    '--linter-output-format',
+    '{filename}:{line}:{code}: {message}',
```
mfussenegger, line 14: The pattern doesn't match this format specification. Are you sure this is working?

Penaz91: I'm pretty sure it's correct, since the `{line}` formatting symbol returns a `line:col` formatted string.

I'm not sure about how the arguments are sent, because the `{filename}:{line}:{code}: {message}` should be a string by itself.

In essence the command would be `djlint --linter-output-format '{filename}:{line}:{code}: {message}'  filename`.

TheEpic-dev:
Hah, I  just added djlint to my `formatter.nvim` config last night and was about to tackle this today. I open GitHub, and lo and behold, it's already done. Thank you!

TheEpic-dev:
Wouldn't this be better implemented using `stdin` though? Then it would work on insert, not just after saving. `djlint` does support `stdin` input if you pass it an empty `-` flag.

```bash
cat /tmp/django.html | djlint --lint - --linter-output-format="{line}:{code}: {message}"
15:15:H025: Tag seems to be an orphan.
15:23:H025: Tag seems to be an orphan.
Linted 1 file, found 2 errors.
```

Penaz91:
> Wouldn't this be better implemented using `stdin` though? Then it would work on insert, not just after saving. `djlint` does support `stdin` input if you pass it an empty `-` flag.
> 
> ```shell
> cat /tmp/django.html | djlint --lint - --linter-output-format="{line}:{code}: {message}"
> 15:15:H025: Tag seems to be an orphan.
> 15:23:H025: Tag seems to be an orphan.
> Linted 1 file, found 2 errors.
> ```

I don't know if it makes much of a difference, mostly due to how nvim-lint works (see https://github.com/mfussenegger/nvim-lint#usage). It seems to be triggered at save with an autocommand. But it is definitely possible to use stdin.

TheEpic-dev:
>  It seems to be triggered at save with an autocommand.

Yup. The difference is that using a file, it can only format on the `BufWritePost` event. With `stdin`, it can format on `BufPostSave`, but also on `InsertLeave` and `TextChanged` events, so you get a quicker feedback loop (no need to save).

I'm also not sure that the pattern in your code was correct. djlint rules look like `D018`, `H025`, etc., so I believe the capture pattern for the code should look like `(%a%d+)`, not just `(%d+)`.

I took the liberty of fixing those two minor things and opened a new PR, #332 :)

PR #330, opened 2023-07-26
adding tfsec lint

mvaldes14:
Including tfsec as a linter. Works as intended on my 4-5 manifests.
Only shows the actual lint messages if the filename matches what the linter is pointing out (prevents random messages showing up in non active files)

Tests added but struggled to make them work, would appreciate some guidance there.
Heavily inspired by how null-ls did it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ Other dedicated linters that are built-in are:
 | [vint][21]                         | `vint`            |
 | [vulture][vulture]                 | `vulture`         |
 | [yamllint][yamllint]               | `yamllint`        |
+| [tfsec][tfsec]                     | `tfsec`           |
 
 ## Custom Linters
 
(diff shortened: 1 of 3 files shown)


mfussenegger:
Added a fixup to make the tests path.

Is the `location.filename` always just the filename? `nvim_buf_get_name` returns the absolute path.

mvaldes14:
Its the absolute path so figured that just giving it the entire nvim_buf_name was sufficient since it has the full path as well.
thanks for fixing the tests!

mfussenegger:
> Its the absolute path so figured that just giving it the entire nvim_buf_name was sufficient since it has the full path as well. thanks for fixing the tests!

If it's the absolute path then my fix was wrong.

I pushed another fixup, can you verify if it still works?

mvaldes14:
it still works as intended, thanks again!

mfussenegger:
Thanks for the PR

PR #348, opened 2023-08-22
fix: shellcheck diagnostics missing the `source` property

chrisgrieser:


chrisgrieser:
test failed cause I forgot the trailing comma. fixed.

PR #349, opened 2023-08-23
Add gdlint

mfussenegger:


aMOPel:
Actually, I noticed that

```lua
local severity_map = {
  ['Error'] = vim.diagnostic.severity.ERROR,
}
```

was the wrong decision. The linter calls everything an "Error', but really they are all just warnings.

So it probably should be:
```lua
local severity_map = {
  ['Error'] = vim.diagnostic.severity.WARN,
}
```

I'm sorry for the inconvenience and thank you for your work!

PR #352, opened 2023-08-26
Handle empty output in `pylint` to avoid JSON decode errors

chrisgrieser:
same issue as in #350, only for `pylint`


mfussenegger:
Thanks

PR #356, opened 2023-08-31
Add lnum_offset and end_lnum_offset options for from_pattern

mfussenegger:


itaranto:
Could these options be added to `from_errorformat` as well?

For consistency sake..

lua/lint/parser.lua
```diff
@@ -39,8 +39,14 @@ function M.from_errorformat(efm, skeleton)
   end
 end
 
+
 --- Parse a linter's output using a Lua pattern
---
+---
+---@param pattern string
+---@param groups string[]
+---@param severity_map table<string, DiagnosticSeverity>
```
itaranto, line 47: I think this should be optional too.

PR #358, opened 2023-09-01
Allow specifying `cwd` per linter

### LINKED ISSUE #338: Allow linters to set `cwd`
Hi there, https://github.com/mfussenegger/nvim-lint/pull/179 introduced `cwd` parameter that can be passed to `try_lint()`. I have added custom linter locally and would really benefit from being able to set `cwd` on linter directly rather than passing it through `try_lint()` which affects all other set up linters.

Do you think it would be possible and sensible to add such functionality?

Thanks.

gegoune:
Closes #338.

Very naive but appears to work fine.


mfussenegger:
What's a setup that requires one linter to have a different cwd than others?

gegoune:
`terraform validate` needs to run in directory of a file rather than repository's root directory. It's common for repository to contain multiple terraform modules and some other linters, like tflint, can run recursively from repository's root.

PR #365, opened 2023-09-14
fix: Handle empty output in `stylelint`

chrisgrieser:
same issue as #350 and #352


mfussenegger:
I think I'll go with https://github.com/mfussenegger/nvim-lint/pull/375 instead

chrisgrieser:
yeah, I was actually thinking of proposing something like that, too.

Makes of course more sense.

PR #372, opened 2023-09-19
fix: use verbose diagnostic codes for `pylint`

chrisgrieser:
Right now, it will report `C0103` as code, instead of the more verbose `invalid-name`. 

Both are equivalent as rule-ids (e.g. to disable in a `pyproject.toml`), but the latter should be preferable because it is more clear what the rule does.


mfussenegger:
See https://github.com/mfussenegger/nvim-lint/pull/279#discussion_r1049721650, this was kinda intentional

chrisgrieser:
hmm yeah, I'd personally prefer the human readable one still, but I kinda see the point with the consistency with the remaining ecosystem

PR #398, opened 2023-10-09
Add cwd arg to fix stream = "both" when linter_cwd concat is needed

fnune:
This call-site of `on_done` was missing `cwd`, breaking instances where concatenating `linter_cwd` to the parser file path was necessary.

This PR forwards that arg so that `stream = "both"` works.


mfussenegger:
Thanks

fnune:
Thank you! Loving the plugin.

PR #404, opened 2023-10-11
feat(snyk): add snyk linter

pbnj:
## TODO 

- [x] add tests
- [x] list tool in readme

### DIFF
--- a/lua/lint/linters/snyk_iac.lua
+++ b/lua/lint/linters/snyk_iac.lua
@@ -1,0 +1,35 @@
+local severity_map = {
+  ["low"] = vim.diagnostic.severity.INFO,
+  ["medium"] = vim.diagnostic.severity.WARN,
+  ["high"] = vim.diagnostic.severity.ERROR,
+}
+
+return {
+  cmd = "snyk",
+  stdin = false,
+  append_fname = true,
+  args = { "iac", "test", "--json" },
+  stream = "stdout",
+  ignore_exitcode = true,
+  env = nil,
+  parser = function(output)
+    local diagnostics = {}
+    local ok, decoded = pcall(vim.json.decode, output)
+    if not ok then
+      return diagnostics
+    end
+    for _, result in ipairs(decoded and decoded.infrastructureAsCodeIssues or {}) do
+      local err = {
+        source = "snyk",
+        message = string.format("%s - %s - %s", result.title, result.issue, result.impact),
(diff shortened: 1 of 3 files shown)


lua/lint/linters/snyk_iac.lua
```diff
@@ -0,0 +1,36 @@
...
+      return diagnostics
+    end
+    for _, result in ipairs(decoded and decoded.infrastructureAsCodeIssues or {}) do
+      local err = {
+        source = "snyk",
+        message = string.format("%s - %s - %s", result.title, result.issue, result.impact),
+        lnum = result.lineNumber - 1,
+        end_lnum = result.lineNumber - 1,
+        col = result.lineNumber,
+        end_col = result.lineNumber,
```
mfussenegger, line 28: Using the lineNumber as columns seems wrong?

pbnj: I copied it from [`tfsec` linter](https://github.com/mfussenegger/nvim-lint/blob/6d596b87862909370f4d1e6535cc9fad80c89fe6/lua/lint/linters/tfsec.lua#L24-L27), but I agree.

If the linter does not provide a column, should I leave `col` and `end_col` out? Or should I set them to `1`?

pbnj: Removed `col` and `end_col`.

mfussenegger: `col` is required according to `:help diagnostic-structure`, so if it isn't provided by the linter it should probably be set to `0`

pbnj: I believe columns start at 1 in (neo)vim.

For example, place the cursor in the 1st column in a buffer and press <kbd>g</kbd> + <kbd>ctrl</kbd><kbd>g</kbd>. Example result: `Col 1 of 105; Line 467 of 48520; Word 1474 of 145763; Byte 49180 of 5216916`

mfussenegger: It uses a mix of indexing. Some is (0,0), some is (1,0). 
Diagnostics are 0 based, from `:help diagnostic-structure`:

> Diagnostics use the same indexing as the rest of the Nvim API (i.e. 0-based
> rows and columns). |api-indexing|

pbnj: I see. Thank you for clarifying this. I will fix this.

PR #409, opened 2023-10-12
fix: support end col for eslint

sQVe:
This PR aims to add end_col support for the `eslint` and `eslint_d` linters. 

See https://github.com/mfussenegger/nvim-lint/issues/394 for background.

PR #416, opened 2023-10-14
fix: repair eslint linters

sQVe:
👋🏼 ,

This PR aims to repair the ESLint issues introduced with https://github.com/mfussenegger/nvim-lint/pull/409, which I apparently didn't validate good enough...

The mocked test JSON should now be correct and the parser correctly picks the first item from the output, which is exactly how `null-ls` has handled it.  

Sorry for the introduced issue... 😮‍💨

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -26,7 +26,10 @@ return require('lint.util').inject_cmd_exe({
   stream = 'stdout',
   ignore_exitcode = true,
   parser = function(output)
-    local success, decodedData = pcall(vim.json.decode, output)
+    local success, decodedData = pcall(
+      vim.json.decode, output,
+      { luanil = { object = true, array = true } }
+    )
     local messages = decodedData and decodedData[1] and decodedData[1].messages or {}
 
     local diagnostics = {}
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -1,7 +1,6 @@
 local severities = {
-  nil,
-  vim.diagnostic.severity.ERROR,
   vim.diagnostic.severity.WARN,
+  vim.diagnostic.severity.ERROR,
 }
 
 return require('lint.util').inject_cmd_exe({
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -36,8 +36,8 @@ return require('lint.util').inject_cmd_exe({
           source = "eslint",
           lnum = diagnostic.line - 1,
           col = diagnostic.column - 1,
-          end_lnum = diagnostic.endLine - 1,
-          end_col = diagnostic.endColumn - 1,
+          end_lnum = (diagnostic.endLine or diagnostic.line) - 1,
+          end_col = (diagnostic.endColumn or diagnostic.column) - 1,
           severity = severities[diagnostic.severity],
           message = diagnostic.message,
           code = diagnostic.ruleId
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -8,9 +8,11 @@ return require('lint.util').inject_cmd_exe({
   cmd = function()
     local local_eslint = vim.fn.fnamemodify('./node_modules/.bin/eslint', ':p')
     local stat = vim.loop.fs_stat(local_eslint)
+
     if stat then
       return local_eslint
     end
+
     return 'eslint'
   end,
   args = {
@@ -25,10 +27,11 @@ return require('lint.util').inject_cmd_exe({
   ignore_exitcode = true,
   parser = function(output)
     local success, decodedData = pcall(vim.json.decode, output)
-    local diagnostics = {}
+    local messages = decodedData and decodedData[1] and decodedData[1].messages or {}
 
-    if success and decodedData ~= nil then
-      for _, diagnostic in ipairs(decodedData.messages or {}) do
+    local diagnostics = {}
+    if success and #messages > 0 then
+      for _, diagnostic in ipairs(messages or {}) do
         table.insert(diagnostics, {
           source = "eslint",
           lnum = diagnostic.line - 1,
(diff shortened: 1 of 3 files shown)


pavkam (approved):

emmanueltouzery:
I think I may have found another regression with the recent eslint changes. With this file:
```javascript
function(x) {
    console.log('test")
}
```

I get:
```
Error executing vim.schedule lua callback: .../pack/packer/start/nvim-lint/lua/lint/linters/eslint.lua:39: attempt to perform arithmetic on field 'endLine' (a nil value)
stack traceback:
	.../pack/packer/start/nvim-lint/lua/lint/linters/eslint.lua:39: in function 'parse'
	...vim/site/pack/packer/start/nvim-lint/lua/lint/parser.lua:131: in function <...vim/site/pack/packer/start/nvim-lint/lua/lint/parser.lua:127>
```

However that PR still improves the state compared to the current main branch.

sQVe:
@emmanueltouzery Are you able to replicate that error with this PR?

emmanueltouzery:
yes, i'm running that PR. you do not reproduce that error?

sQVe:
@emmanueltouzery Ah, I had the wrong branch checked out in my dev directory. I will fix this and add it to this PR.

Cheers for pointing out the issue 🍻

sQVe:
@emmanueltouzery The issue should now be fixed, and it should give you an ` eslint_d: Parsing error: Unterminated string constant ` instead of erroring. If you can, I would appreciate it if you could verify 🙏🏼

emmanueltouzery:
yes, the newest commit fixes the issue. The message is a little strange though, the error code is vim.NIL?

Diagnostics:
1. typescript: Unterminated string literal. [1002]
2. eslint: Parsing error: Unterminated string literal. [vim.NIL]

But this branch is way better than the nvim-lint main branch!

sQVe:
> 2. eslint: Parsing error: Unterminated string literal. [vim.NIL]

Yeah, this is due to the JSON decode in Neovim. I'll see if I can make this look nicer later today. Hopefully it's as easy as making the decoder return `nil` instead.

sQVe:
@emmanueltouzery This `vim.NIL` issue should now be fixed with the latest commit.

lua/lint/linters/eslint.lua
```diff
@@ -24,17 +26,21 @@ return require('lint.util').inject_cmd_exe({
...
-    local diagnostics = {}
+    local success, decodedData = pcall(
+      vim.json.decode, output,
+      { luanil = { object = true, array = true } }
+    )
+    local messages = decodedData and decodedData[1] and decodedData[1].messages or {}
 
-    if success and decodedData ~= nil then
-      for _, diagnostic in ipairs(decodedData.messages or {}) do
+    local diagnostics = {}
```
pavkam, line 35: I wonder about the severities. In my local config I had to update `severties` because they did not match the actual ones:

```
    "off" or 0 - turn the rule off
    "warn" or 1 - turn the rule on as a warning (doesn’t affect exit code)
    "error" or 2 - turn the rule on as an error (exit code is 1 when triggered)
```
from https://eslint.org/docs/latest/use/configure/rules

sQVe: 👍🏼 

This should now be fixed.

pavkam:
Can we get this in?

AlariCode:
Merge please) Can’t switch from nullls.

mfussenegger:
Have some patience. This isn't a paid fulltime job, you don't have a service contract and you can always use branches or pin to older commits.

kasdeya:
I'm getting this error using this commit when I open my eslint config  (.eslintrc.cjs) or my tailwind config (tailwind.config.js).

Error executing vim.schedule lua callback: .../share/nvim/lazy/nvim-lint/lua/lint/linters/eslint_d.lua:39: attempt to perform arithmetic on field 'line' (a nil value)
stack traceback:
	.../share/nvim/lazy/nvim-lint/lua/lint/linters/eslint_d.lua:39: in function 'parse'
	.../.local/share/nvim/lazy/nvim-lint/lua/lint/parser.lua:131: in function <.../.local/share/nvim/lazy/nvim-lint/lua/lint/parser.lua:127>

sQVe:
@kasdeya Sign, that's disappointing. It seems that the output of `eslint` isn't reliable at all. I don't get how `line` could be missing from a diagnostic output.

@mfussenegger Do you have any thoughts on how we can guard against these issues?

kasdeya:
@sQVe including my .eslintrc.cjs file in my tsconfig.json file seems to make the error stop, but then im linting that file, and this wasn't happening before, for now I added a check for the nil value to just return the empty table, seems to work for me so far, but yeah i don't know, this is me trying on local projects I had lint working before.

mfussenegger:
> Do you have any thoughts on how we can guard against these issues?

Skipping entries where the line is missing could be an option,  but it would be good to know how the exact output looks like. 

I could merge https://github.com/mfussenegger/nvim-lint/pull/419 to get back to the known working state in the meantime to put the fire out,  unless someone provides another fix

sQVe:
@mfussenegger Please do. I'll open a new PR that fixes all this once and for all.

@kasdeya Can you provide the output of `eslint --format json` for the file you're trying to lint? I'm curious why you're not getting any line value.

kasdeya:
I just added this on the for loop that builds the diagnostics table and it works for me, lints in my project files and doesn't throw an error when opening my .eslintrc.cjs in my project.

```lua
    local diagnostics = {}
    if success and #messages > 0 then
      for _, diagnostic in ipairs(messages or {}) do
        if diagnostic.line == nil then
          return diagnostics
        end
        ...
```

sQVe:
@kasdeya But how would it display a diagnostic without the `line` value?

kasdeya:
@sQVe @mfussenegger #421 I made a pull request with a quick fix. Figured some errors might not contain a line number, this happened when .eslintrc.cjs was not included in tsconfig.json for me, I don't want to include it cause then it will give me a ton of lint errors in that file, same was happening with my tailwind config file. This displays that message in the first line, instead of throwing an error whenever you open the file or run try_lint(), let me know what you think, this replicates the behavior that nvim-lint used to have.

Issue #357, opened 2023-08-31
Long running linters can accumulate


TimDeve:
It seems there is a bug in the part of codebase that ensures that when a linter is run it sigterms the previous invocation of it, it sometimes does but most of the time it seems to accumulate. 

My gut says that there might be a race around lua/lint.lua:154 and that storing previous linter in a pid - > proc table might avoid it but I've not had the time to validate that.

OldWorldOrdr:
any progress on this? my system has froze multiple times when poking around in autoconf configure scripts

TimDeve:
It seems that `lint` and `try_lint` erase each other's data. [Here](https://github.com/mfussenegger/nvim-lint/compare/master...TimDeve:nvim-lint:sigterm-old-lints)'s a basic diff that fixes it. I've not had the time to figure out why `lint` did that cleanup in the first place though so I've not raised a PR yet.

TimDeve:
#417 has been merged which fixes the issue

PR #417, opened 2023-10-15
Fixes race causing previous processes to not be cleaned up

TimDeve:
There can be a race between the `on_exit` callback of `uv.spawn` and `try_lint` setting `procs[linter.name]` meaning that long running processes may not be cleaned up properly as described in #357.

This PR adds a check that ensures that we only cleanup `running_procs_by_buf` when the handle that triggered the callback is same that is in `procs[linter.name]`.

This technically doesn't stops the callback and `try_lint` racing, it just ensure the callback doesn't remove handles it shouldn't. When the last process in a chain finishes and find it's handle in the data structure then it cleans up.

I have tested this PR using a linter like so, that enabled me to observe long running processes and `:w` several time in sequence to observe the bug and its fix.
```
  require('lint').linters.longsleep = {
    cmd = 'bash',
    stdin = false,
    append_fname = false,
    args = {"-c", "sleep 100"},
    stream = "stdout",
    ignore_exitcode = true,
    parser = function(output, bufNo) return {} end
  }
```

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -218,9 +218,12 @@ function M.lint(linter, opts)
   handle, pid_or_err = uv.spawn(cmd, linter_opts, function(code)
     if handle and not handle:is_closing() then
       local procs = (running_procs_by_buf[bufnr] or {})
-      procs[linter.name] = nil
-      if not next(procs) then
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks

PR #421, opened 2023-10-17
Eslint d fix error without line number

kasdeya:
This is a quick fix for eslint_d returning diagnostics without line numbers.  #416 
sorry for the formatting changes, this fixes the error I was having, I think the error was that sometimes eslint_d will return errors without line number, this will print that error in the first line of the file as a warning like it used to happen on previous versions, let me know what you think.

### DIFF
--- a/lua/lint/linters/eslint_d.lua
+++ b/lua/lint/linters/eslint_d.lua
@@ -46,18 +46,18 @@ return require("lint.util").inject_cmd_exe({
             code = diagnostic.ruleId,
           })
           return diagnostics
+        else
+          table.insert(diagnostics, {
+            source = "eslint_d",
+            lnum = diagnostic.line - 1,
+            col = diagnostic.column - 1,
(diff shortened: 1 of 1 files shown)


sQVe:
I think it would be better if we properly validated the  diagnostics rather than doing on off checks like this. Since the `eslint` output isn't reliable, I wouldn't be surprised if it could return `nil` for more values than `line`.

kasdeya:
@sQVe  from what I saw eslint_d will always return a message, a severity and a code if there is a diagnostic, only the line number could be nil, since if there's no diagnostic message then there's no diagnostic from what I understand.

kasdeya:
I think an if else statement might work better though, just in case there's both, messages without line numbers and with line numbers in the same diagnostic.

sQVe:
@kasdeya I just pushed https://github.com/mfussenegger/nvim-lint/pull/423, could you give it a glance?

I think it's more robust and lenient, with a bunch of new tests to verify.

kasdeya:
closing this now in favor of #423

PR #423, opened 2023-10-17
fix: parse eslint output with json

sQVe:
This PR aims to parse `eslint` JSON output, which will improve the diagnostics since it allows you to get more information, like `end_col`.

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -1,30 +1,13 @@
 local binary_name = "eslint"
-
 local severities = {
   vim.diagnostic.severity.WARN,
   vim.diagnostic.severity.ERROR,
 }
 
-local diagnostic_translate_map = {
-  column = "col",
-  endColumn = "end_col",
-  endLine = "end_lnum",
-  line = "lnum",
-  message = "message",
-  ruleId = "code",
-  severity = "severity",
-}
-
 return require('lint.util').inject_cmd_exe({
   cmd = function()
     local local_binary = vim.fn.fnamemodify('./node_modules/.bin/' .. binary_name, ':p')
-    local stat = vim.loop.fs_stat(local_binary)
-
-    if stat then
-      return local_binary
-    end
-
-    return binary_name
+    return vim.loop.fs_stat(local_binary) and local_binary or binary_name
   end,
   args = {
     '--format',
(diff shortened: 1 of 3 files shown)


lua/lint/linters/eslint.lua
```diff
@@ -1,27 +1,74 @@
-local pattern = [[%s*(%d+):(%d+)%s+(%w+)%s+(.+%S)%s+(%S+)]]
-local groups = { 'lnum', 'col', 'severity', 'message', 'code' }
-local severity_map = {
-  ['error'] = vim.diagnostic.severity.ERROR,
-  ['warn'] = vim.diagnostic.severity.WARN,
-  ['warning'] = vim.diagnostic.severity.WARN,
+local binary_name = "eslint"
```
sQVe, line 1: I added this since the `eslint` and `eslint_d` linters are mirrors. We could set this up differently to keep things more DRY, but yeah... 🤷🏼

lua/lint/linters/eslint.lua
```diff
@@ -1,27 +1,74 @@
...
+          if json_diagnostic[json_key] ~= vim.NIL then
+            if json_key == 'severity' then
+              diagnostic[diagnostic_key] = severities[json_diagnostic[json_key]]
+            elseif type(json_diagnostic[json_key]) == "number" then
+              diagnostic[diagnostic_key] = json_diagnostic[json_key] - 1
+            else
+              diagnostic[diagnostic_key] = json_diagnostic[json_key]
+            end
+          end
+        end
```
sQVe, line 59: We're now looking up a specific set of JSON keys, and only adding them if they aren't `vim.NIL`. This should remove all `nil` issues we've had previously.

sQVe: This is fairly similar to how `null-ls` handled this. See https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/helpers/diagnostics.lua#L184-L188.

lua/lint/linters/eslint.lua
```diff
@@ -1,27 +1,74 @@
...
+
+        -- Ensure that we have a diagnostic with `col` and `lnum` set, since
+        -- they are required.
+        if diagnostic.col and diagnostic.lnum then
```
sQVe, line 63: Only add the diagnostic if the `col` and `lnum` is available, since they are required.

sQVe (commented):
I've added more extensive tests for this, where it checks all the issues we've encountered so far.

sQVe:
@mfussenegger Give this a glance when you have the time. ❤️

kasdeya:
This does not throw the error, however it doesn't show the errors that have no line number, clone this repo and open the .eslintrc.cjs file (https://github.com/kasdeya/lint-test), try it with my commit and it'll show the eslint message that had no line number, on line one, I think in some cases it may be important to show that error maybe...

kasdeya:
Let me know what you think about showing that warning, it could be useful in some cases maybe? @sQVe

sQVe:
@kasdeya I think it's weird the way they've set this up. It seems like a missing column and/or line is a way to give an error for the whole file.

I would agree that it could be really useful to show the error in this case.

This is what the error output looks like:
```
  0:0  error  Parsing error: ESLint was configured to run on `<tsconfigRootDir>/.eslintrc.cjs` using `parserOptions.project`: <tsconfigRootDir>/tsconfig.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file
```

I'll see how we can best support this.

kasdeya:
You can just add that diagnostic to the first line too, that's what I did with my solution, it has no line number, because it pertains to the whole file I assume, not a line itself, my solution might be simple but it gets the job done heh

sQVe:
@kasdeya The latest commits should add support for handling `fatal` diagnostics. See https://github.com/mfussenegger/nvim-lint/pull/423/commits/a55f72fe5a33ff3e3048715d9786e8c89a947c62.

kasdeya:
@sQVe  Looks good to me!

mfussenegger:
@sQVe this felt a tad bit over-engineered to me so I pushed a fixup to change things around a bit. Can you verify that it still works?

PR #425, opened 2023-10-18
Introduce optional bufnr into opts to try_lint and lint

pavkam:
Helps when trying to lint a non-active buffer.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -128,7 +128,7 @@ function M.try_lint(names, opts)
     vim.diagnostic,
     "nvim-lint requires neovim 0.6.0+. If you're using an older version, use the `nvim-05` tag of nvim-lint'"
   )
-  opts = opts or {}
+  opts = vim.tbl_extend("keep", opts or {}, { bufnr = api.nvim_get_current_buf() })
   if type(names) == "string" then
     names = { names }
   end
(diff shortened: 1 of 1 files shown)


mfussenegger:
You could use `nvim_buf_call`. Linting non-active buffers seems rather niche to me. Why would you do that?

pavkam:
> You could use `nvim_buf_call`. Linting non-active buffers seems rather niche to me. Why would you do that?

That's what I used but it seems like a last-ditch-effort when other options are not available. Since this is a small enough change seemed like a good idea to have it in the code.

In my case, I turn on/off linting on buffer/global basis due to the size of my project. When I turn it off I remove all diagnostics generated by `nvim-lint`. When I turn it on I want to re-lint the open buffers to check all is fine. That's why I need to run it in a different buffer. Maybe it's a niche thing, but I find it useful.

mfussenegger:
> In my case, I turn on/off linting on buffer/global basis due to the size of my project. When I turn it off I remove all diagnostics generated by nvim-lint. When I turn it on I want to re-lint the open buffers to check all is fine. That's why I need to run it in a different buffer. Maybe it's a niche thing, but I find it useful.

I see, makes sense.

> Since this is a small enough change seemed like a good idea to have it in the code.

Unfortunately it isn't that small. Some linters use functions in their properties and assume that these functions are executed under the current buffer:

https://github.com/mfussenegger/nvim-lint/blob/7c8b50a2bb62b544eb91434fd008a34c4b51fd6d/lua/lint/linters/ruff.lua#L4-L18


The exception is the `parse` function, which gets the `bufnr` already as argument because of the async execution of the linter.

The built-in linters could be adopted, but it would be a breaking change for any custom linters currently relying on this behavior.

Because of that I'd rather defer to `nvim_buf_call`. At least for now. If this comes up more often I might reconsider.

PR #431, opened 2023-10-20
Correct cmakelint cmd name (cmake-lint)

konosubakonoakua:


mfussenegger:
Are there different versions of cmakelint? The one I could find has a `cmakelint` binary.

konosubakonoakua:
turns out that it is a lazyvim cmakelang plugin config problem, PR can be closed, thanks.

PR #432, opened 2023-10-22
fix: support end col for mypy

landeaux:
This PR adds end column support for the `mypy` linter.

Also, there was no existing test coverage for this linter, so I added it. Tests are passing locally.


mfussenegger:
Thanks

PR #437, opened 2023-10-27
sqlfluff: Remove 'None' lines from output

jpfender:
sqlfluff occasionally exhibits weird behaviour where it will break in the middle of linting and display a line that just reads `None` instead of showing JSON output:

sqlfluff_cli

It will then continue linting and eventually print the proper JSON output. However, the initial `None` line becomes part of the `output` string in the parser, which leads to a `jsondecodeerror` and makes nvim-lint just display the raw output on line 1 of the buffer:

sqlfluff_nvim_incorrect

If we strip the `None` line from the output, the diagnostic is displayed correctly:

sqlfluff_nvim_correct

### DIFF
--- a/lua/lint/linters/sqlfluff.lua
+++ b/lua/lint/linters/sqlfluff.lua
@@ -10,6 +10,7 @@ return {
   parser = function(output, _)
     local per_filepath = {}
     if #output > 0 then
+      output = output:gsub("None\n", "")
       local status, decoded = pcall(vim.json.decode, output)
       if not status then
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
This seems a bit like a workaround for an upstream bug.

Wouldn't it make more sense to figure out what's causing `None` to be printed in sqlfluff and fix it there?

mfussenegger:
Closing this due to lack of response. Please open again if this is still relevant.

PR #438, opened 2023-10-27
ignore lint results from processes we kill

### LINKED ISSUE #388: cancelling processes with sigterm resets the diagnostics
the idea of killing existing linter processes when a new concurrent lint is requested is causing the diagnostics to reset for me.

at least in the case of eslint, when we call `proc:kill("sigterm")`, nvim-lint thinks that the eslint that we just killed finished successfully and found that there are no linting errors. Therefore all diagnostics are removed.
A little later, the new-running linter succeeds and now diagnostics are added back.

On my setup, due to the signs column, it makes the code shift left (diagnostics are removed) then back right (they are added back).

I'm not sure whether it would be possible to detect that the linter failed (was killed by sigterm), and that therefore we should not reset the diagnostics, to avoid this issue.

I confirmed that this change fully solves the issue for me:
```diff
  for linter_name, proc in pairs(running_procs) do
-   if not proc:is_closing() then
-     proc:kill("sigterm")
-   end
-   running_procs[linter_name] = nil
+   return
  end
```

So from my point of view, I see two solutions:
1. instead of killing existing running processes, let the existing one finish and don't run the new one (that still solves the problem of having many concurrently running processes) -- this is what my change does
2. somehow detect that the process was killed by nvim-lint and don't reset the diagnostics to empty in that case

emmanueltouzery:
fixes #388
when we want to run a new lint process, but an existing one is already running, we kill the existing process. That process then exits immediately. Until now we took into account the results from that linter run and reported them to the user. Better to ignore this incomplete run, to avoid a "blinking effect" for the user, with errors disappearing and then immediately returning.
This can particularly happen with slow linters (for instance eslint, which can easily run for one second).

An easy way to reproduce the issue is to open a file with linter errors and using a slow linter (possibly faking a slow linter by introducing a sleep call), then triggering linting twice in a row quickly. When we trigger the second time, the first linter is killed, and the errors "disappear", only to reappear when the second linter completes. With this change, the errors do not disappear anymore, they just stay present the whole time.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -60,20 +60,29 @@ local namespaces = setmetatable({}, {
   end
 })
 
+local last_killed_pid = nil
 
-local function read_output(cwd, bufnr, parser, publish_fn)
+local function read_output(pid_or_err, cwd, bufnr, parser, publish_fn)
   return function(err, chunk)
     assert(not err, err)
     if chunk then
       parser.on_chunk(chunk, bufnr)
     else
-      parser.on_done(publish_fn, bufnr, cwd)
+      if last_killed_pid ~= pid_or_err then
+        parser.on_done(publish_fn, bufnr, cwd)
+      else
+        -- We arrived here because we killed the process when we wanted to launch
+        -- a new linter. Killing the process caused it to return immediately, most
+        -- likely without errors (or with incomplete errors). Ignore this run, we'll
+        -- rather collect the full data from the next run.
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks, but I went with a bit of a different solution to avoid the extra global state: https://github.com/mfussenegger/nvim-lint/pull/442/commits/e433148f97073d564c7c151d82e0e109093ca15a

PR #444, opened 2023-11-03
feat(linter): Added pyproject-flake8 support

toirl:
Hi,

I love flake8, and I even more I love configuring it using `pyroject.toml` files, which isn't supported by flake8. However there is a package (https://github.com/csachs/pyproject-flake8) which makes this possible. 

I currently run nvim-lint with a custom configuration for this linter which works well. So I thought it might be a nice idea to add this to upstream also.


mfussenegger (commented):
Could you please revert the unrelated style changes?

Is there any reason to use this over https://docs.astral.sh/ruff/ - which is faster and also supports pyproject.toml?

toirl:
Hi, just reverted the changes on style. 

Regarding the question for reasons against ruff: No reasons against ruff of course. I just did not used it yet. The often mentioned faster speed of ruff wasn't  valuable enough for me to change my current setup. Never had issues with the speed of the other linters.

PR #447, opened 2023-11-04
ruff: Implement json parsing (for end_col) and basic code to severity mapping

Avishayy:
JSON parsing for end_col (to get underlinesj

In addition, I disliked having warnings on all lints, so I copied the mapping from [ruff-lsp](https://github.com/astral-sh/ruff-lsp/blob/6a1a7b18711cf65b058492d5800e606f14556861/ruff_lsp/server.py#L346-L354)


aignas:
Just wanted to say that I've tested this and it is really nice! Thank you!

aignas (approved):

mfussenegger:
Thanks. I amended a small change and added the source name too.

PR #450, opened 2023-11-05
fix(ruff): do not ignore the exit code

aignas:
With this fix the `ruff` diagnostics get shown on my neovim screen.


aignas:
It seems that for others it is working as intended. Marking as draft whilst I am investigating.

aignas:
Sorry for the noise, everything is working as expected.

PR #474, opened 2023-11-23
docs: add `typos` linter

chrisgrieser:
`typos` already has an existing config, but was missing from the docs.


chrisgrieser:
made another commit that fixes the issue of file-exclusions in `typos.toml` not being respected.

PR #475, opened 2023-11-25
Add support for `prisma-lint`

maxh:
I recently added a JSON output format to `prisma-lint` to support editor integrations: https://github.com/loop-payments/prisma-lint

image

### DIFF
--- a/lua/lint/linters/prisma_lint.lua
+++ b/lua/lint/linters/prisma_lint.lua
@@ -1,0 +1,35 @@
+local binary_name = "prisma-lint"
+
+return require('lint.util').inject_cmd_exe({
+  cmd = function()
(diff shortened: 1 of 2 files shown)


lua/lint/linters/prisma_lint.lua
```diff
@@ -0,0 +1,38 @@
+local binary_name = "prisma-lint"
```
maxh, line 1: Would it be better to call this file `prisma-lint` (with a hyphen instead of an underscore) instead to match? Not sure what's idiomatic here.

mfussenegger: So far the files tend to use underscore if the binary has a hyphen.

Could you also add the linter to the readme?

maxh:
Rebased master and updated to account for https://github.com/mfussenegger/nvim-lint/pull/476

PR #483, opened 2023-12-05
Allow getting the running linters for a specified buffer

mehalter:
This allows the new `get_running` function to be provided a specific buffer number to retrieve linters for

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -171,12 +171,15 @@ end
 local running_procs_by_buf = {}
 
 
---- Returns the names of the running linters
+--- Returns the names of the running linters for a given buffer
+---@param bufnr? integer buffer number to get the linters from, defaults to current
 ---
 ---@return string[]
-function M.get_running()
+function M.get_running(bufnr)
   local linters = {}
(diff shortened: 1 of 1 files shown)


mehalter:
Just to add some more context, this is quite useful when developing stuff like `statusline` and also enables the ability to configure window specific `statusline` for people who do not use global `statusline`.

I have also made sure that the change is non-breaking with the already implemented API for getting the running linters. It simply adds the ability to check other buffers as well.

mfussenegger:
> I have also made sure that the change is non-breaking with the already implemented API for getting the running linters

Good call, but I think in this case I'd make a slight behaviour change: Return all active linter name if bufnr is nil and only use the current buffer if it's 0
Otherwise that would require a new function, or another parameter to serve that use case - both not ideal.

mfussenegger:
Thanks for the PR. I merged https://github.com/mfussenegger/nvim-lint/pull/489 instead

mehalter:
Awesome! That's totally fine! Sorry I didn't update here quicker. Super appreciate the added functionality tho :)

PR #488, opened 2023-12-11
Support finding eslint in root node_modules

marlun:
When working in a workspace there might be multiple node_modules folders and eslint might not be in the closest one. User find() to search multiple folders

I'm not sure about the limit being set to 3 but it seemed like a small enough number which would work for most repos. That's a guess though.

I haven't been able to run the tests yet because busted is complaining about `attempt to index a nil value (global 'vim')`. I've read some about running it through neovim but haven't understood it yet.

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -4,10 +4,21 @@ local severities = {
   vim.diagnostic.severity.ERROR,
 }
 
+local function get_local_binary()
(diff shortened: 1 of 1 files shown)


mfussenegger:
Sorry but no, if the current default implementation doesn't work you'll have to customize the linter config.

marlun:
No worries, will look into customizing the config.

PR #493, opened 2023-12-23
feat: add dash linter

bydlw98:
Uses `dash -n` to check syntax of shell scripts without executing them.
This is similar to `zsh --no-exec` and `fish --no-execute`.


mfussenegger:
Thanks

PR #494, opened 2023-12-23
Add linter for pony using the ponyc compiler

sacovo:


mfussenegger (commented):
Thanks for the PR. Could you also add the linter to the readme?

Issue #495, opened 2023-12-24
All parsers relying on `from_pattern` are broken


stephen-huan:
3739681 introduces a new parameter `bufnr` that only displays the lint if the item buffer matches `bufnr`. This appears to breaks every parser relying on the `require("lint.parser").from_pattern` pattern.

I personally bisected the error for the fish and statix linters. This is likely also the root cause of #481 and #492.

mfussenegger:
I don't see how a change to `from_errorformat` would break `from_pattern` and several linters that use `from_pattern` work just fine for me.

Given that you didn't post any reproduction steps, or provided an actual error description I'm closing this.

stephen-huan:
@mfussenegger sorry, I overgeneralized too soon. for fish, the file

```fish
echo "$$"
```

`fish --no-execute` gives

```
temp.fish (line 1): $$ is not the pid. In fish, please use $fish_pid.
echo "$$"
       ^
warning: Error while reading file temp.fish
```

and indeed, the editor diagnostic shows

```
Diagnostics:
1. $$ is not the pid. In fish, please use $fish_pid.
```

as expected on commit 6f589cb93560581dc2a3b9693658afe865e5649e. On commit 373968190b0022be80a6099ecb483cfc2b5fe8af, it shows nothing. That is what I meant by bisection.

For statix, the file

```file
{
  x = x;
}
``` 
with `statix check` gives

```
[W03] Warning: Assignment instead of inherit
   ╭─[temp.nix:2:3]
   │
 2 │   x = x;
   ·   ───┬──
   ·      ╰──── This assignment is better written with inherit
───╯
```

and the editor shows 

```
This assignment is better written with `inherit`
```

Like fish, it shows the lint on the working commit, and fails to show on the next commit.

The thing that these linters have in common is they both use `.from_pattern`. I saw the two open issues (I didn't personally bisect them) also used `from_pattern` and generalized too soon.

stephen-huan:
#481 (php) also bisects to 373968190b0022be80a6099ecb483cfc2b5fe8af.

Consider the example used in the original issue

```php
<?php

echo "Hello, world!" // Notice the missing semi-colon at the end of this line
```

Running `php -l` gives

```text
Parse error: syntax error, unexpected end of file, expecting "," or ";" in temp.php on line 4
Errors parsing temp.php
```

and the editor shows

```text
Diagnostics:
1. syntax error, unexpected end of file, expecting "," or ";"
```

[...]

PR #498, opened 2023-12-28
fix(lua/lint/parser.lua): `from_errorformat` and `stdin = true` parsers

### LINKED ISSUE #481: PHP linter not working
When using `php` in the linters with a `*.php` file, the `require("lint").try_lint()` does not display anything when there is an obvious PHP error as follow.

```php
<?php

echo "Hello, world!" // Notice the missing semi-colon at the end of this line
```

When running `php` in the command line manually with the options from this file https://github.com/mfussenegger/nvim-lint/blob/f20f35756e74b91c0b4340d01fee22422bdffefa/lua/lint/linters/php.lua#L11-L14 it gives the correct and expected errors. I'm not advanced enough using Lua with Neovim to understand what is going on. Here is my configuration for `nvim-lint`.

```lua
return {
  "mfussenegger/nvim-lint",
  dependencies = {
    "WhoIsSethDaniel/mason-tool-installer.nvim",
  },
  event = { "BufReadPre", "BufWritePost" },
  init = function()
    local lint = require("lint")

    vim.api.nvim_create_autocmd(
    { "TextChanged", "TextChangedI", "BufEnter", "BufReadPre", "BufWritePost", "CursorHold", "CursorHoldI" },
    {
      callback = function()
        lint.try_lint()
      end,
    }
    )
  end,
  config = function()
    local lint = require("lint")
    local phpcs = require("lint").linters.phpcs

    phpcs.args = {
      "-q",
      -- <- Add a new parameter here
      "--standard=PSR12",
      "--report=json",
      "-",
    }

    lint.linters_by_ft = {
      lua = { "luacheck" },
      javascript = { "eslint" },
      typescript = { "eslint" },
      javascriptreact = { "eslint" },
      typescriptreact = { "eslint" },
      vue = { "eslint" },
      sh = { "shellcheck" },
      fish = { "fish" },
      json = { "jsonlint" },
      markdown = { "markdownlint" },
      php = { "php", "phpcs" },
      css = { "stylelint" },
      yaml = { "yamllint" },
    }
  end,
}

```

As a side note, and after opening this issue, using PHPStan (which does the same as `php -l` plus some useful checkings) does work as intended and displays the expected PHP error about the missing semi-colon.

image

stephen-huan:
373968190b0022be80a6099ecb483cfc2b5fe8af introduces a new parameter `bufnr` that only displays the lint if the item buffer matches `bufnr`. However, parsers that read their input from stdin (`stdin = true`) either do not report a buffer (in which case `item.bufnr` is 0) or incorrectly report their buffer as some variant of `stdin`, `<stdin>`, etc. (when the filename used to be parsed with `%f`).

According to `:help getqflist()`, an item with without a buffer will have `item.bufnr` be 0. 

> Quickfix list entries with a non-existing buffer number are returned
> with "bufnr" set to zero (Note: some functions accept buffer number zero
> for the alternate buffer, you may need to explicitly check for zero).

In this case, we should show the lints. 

Additionally, certain linters (credo, statix) incorrectly parse the filename with `%f` . This has also been fixed.

Fixes #481, #495. Also fixes
- blocklint
- credo
- php
- statix

Fish is out of scope and is actually caused by an [upstream issue](https://github.com/fish-shell/fish-shell/issues/10171).

### DIFF
--- a/lua/lint/linters/credo.lua
+++ b/lua/lint/linters/credo.lua
@@ -1,4 +1,4 @@
-local errorfmt = '[%t] %. %f:%l:%c %m, [%t] %. %f:%l %m'
+local errorfmt = '[%t] %. stdin:%l:%c %m, [%t] %. stdin:%l %m'
 
 return {
   cmd = 'mix',
@@ -8,5 +8,3 @@ return {
   ignore_exitcode = true, -- credo only returns 0 if there are no errors
   parser = require('lint.parser').from_errorformat(errorfmt, { ['source'] = 'credo' })
 }
-
-
--- a/lua/lint/linters/statix.lua
+++ b/lua/lint/linters/statix.lua
@@ -4,7 +4,7 @@ return {
   args = {'check', '-o', 'errfmt', '--stdin'},
   stream = 'stdout',
   ignore_exitcode = true, -- statix only returns 0 if there are no errors
-  parser = require('lint.parser').from_errorformat('%f>%l:%c:%t:%n:%m', {
+  parser = require('lint.parser').from_errorformat('<stdin>>%l:%c:%t:%n:%m', {
     source = 'statix'
   })
 }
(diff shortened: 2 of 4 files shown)


mfussenegger:
Thanks

PR #504, opened 2024-01-06
Add regal and opa_check linters for rego files

FollieHiyuki:
Regal recommends users to run `opa check --strict` alongside it (regal will only take effect when `opa check` passes), so I think it's fine including both of them in the same PR here.

Ref: https://docs.styra.com/regal#opa-check-and-strict-mode


mfussenegger:
Thanks. Merged via https://github.com/mfussenegger/nvim-lint/pull/512 as I couldn't resolve the conflicts on this PR because it was created from the master branch. Please always use a feature branch when creating PRs

PR #510, opened 2024-01-18
Fix tool key name

mustafaergul:
Seems like there is a typo for `erb_lint` tool to make it work. That should be `erb_lint` instead `erb-lint` while using inside lua configuration

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ Other dedicated linters that are built-in are:
 | [djlint][djlint]                   | `djlint`               |
 | [dotenv-linter][dotenv-linter]     | `dotenv_linter`        |
 | [editorconfig-checker][ec]         | `editorconfig-checker` |
-| [erb-lint][erb-lint]               | `erb_lint`             |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,7 +109,7 @@ Other dedicated linters that are built-in are:
 | [djlint][djlint]                   | `djlint`               |
 | [dotenv-linter][dotenv-linter]     | `dotenv_linter`        |
 | [editorconfig-checker][ec]         | `editorconfig-checker` |
-| [erb-lint][erb-lint]               | `erb_lint`             |
+| [erb_lint][erb_lint]               | `erb_lint`             |
```
mfussenegger, line 112: The link definition is still erb-lint, see https://github.com/mfussenegger/nvim-lint/blob/d44185d3e57e00c164485a454174294b36ef7067/README.md?plain=1#L449

mustafaergul: I guess I was confused with the formatted I use, you can ignore this one.

PR #513, opened 2024-01-19
fix: normalize paths before comparing equality

### LINKED ISSUE #480: Backslash in the Windows path causes missing error report
Thanks for this nice plugin.

When used on the Windows platform, I noticed that the backslash in the Windows path caused missing error report. The corresponding code is here: https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/parser.lua#L72

I think we can normalize the path by replacing all backslashes with slashes. I tested with the following code and it worked:

```lua
      local normalized_path = string.gsub(path, '\\', '/')
      local normalized_buffer_path = string.gsub(buffer_path, '\\', '/')
      if normalized_path ~= normalized_buffer_path then
        return nil
      end
```

replicaJunction:
Fix consistency issues on Windows when some linters use a single backslash, some use a double (escaped) backslash, and some still use a forward slash.

This fix is confirmed to fix issues with Mypy on Windows, and it probably affects other linters as well.

Fixes #480.

### DIFF
--- a/lua/lint/parser.lua
+++ b/lua/lint/parser.lua
@@ -69,7 +69,9 @@ function M.from_pattern(pattern, groups, severity_map, defaults, opts)
       else
         path = vim.fn.simplify(linter_cwd .. '/' .. captures.file)
       end
-      if path ~= buffer_path then
+      local normalized_path = vim.fs.normalize(path)
+      local normalized_buffer_path = vim.fs.normalize(buffer_path)
+      if normalized_path ~= normalized_buffer_path then
         return nil
       end
     end


replicaJunction:
Although this is a small change, it makes a big difference in usability for me. Is there anything I can do to get this ready for review?

mfussenegger:
Thanks. Added a small fixup to maintain compatibility with older neovim versions and merged via https://github.com/mfussenegger/nvim-lint/pull/539

For the future please use a feature branch. I cannot ammend to PRs created from master.

replicaJunction:
Understood. I apologize - I thought forking the repo counted my master as being a feature.

I appreciate you looking at this and merging the fix. Thank you!

PR #518, opened 2024-01-22
Use `stderr` for `stylelint`

### LINKED ISSUE #500: stylelint doesn't lint
CleanShot 2024-01-02 at 13 29 53

I don't get any diagnostics for existing error

Seems like `stdout.read_start` callback is called once https://github.com/mfussenegger/nvim-lint/blob/4dbc7ec60b33b656f7c54bb945671a55b18699f2/lua/lint.lua#L147

and here https://github.com/mfussenegger/nvim-lint/blob/4dbc7ec60b33b656f7c54bb945671a55b18699f2/lua/lint.lua#L77 both args `err, chunk` are `nil`

My config https://github.com/kapral18/config.nvim/blob/dee5f388194a315bfbf7d8587b7d3ce8de9dbe60/lua/plugins/css.lua#L42 (lazyvim based)

Nvim: nvim 0.10.0

jmpolom:
This seems to and should resolve #500

### DIFF
--- a/lua/lint/linters/stylelint.lua
+++ b/lua/lint/linters/stylelint.lua
@@ -22,7 +22,7 @@ return {
       return vim.fn.expand("%:p")
     end,
   },
-  stream = "stdout",
+  stream = "stderr",
   ignore_exitcode = true,
   parser = function (output)
     local status, decoded = pcall(vim.json.decode, output)


mfussenegger:
@cpof-tea can you double check this? You made the last changes to this linter, and I don't use it.

jmpolom:
Looks like at some point stylelint changed from outputting problems to `stdout` to `stderr`: https://github.com/stylelint/stylelint/blob/main/docs/migration-guide/to-16.md#breaking-changes

cpof-tea:
@jmpolom indeed. I have stylelint v14.15.0, and it outputs to `stdout`

@mfussenegger is there any way we can dynamically change `stream` property for backward compatibility?

mfussenegger:
If the linter doesn't output anything in the other stream, setting `both` could be an option.

Otherwise the linter could be implemented as `function()` that first calls `stylelint --version` (if that's supported), parses the version and depending on that returns a different table. But not sure if that's worth it.

jmpolom:
`stylelint -v` does return a version string. Would it be possible to define the `stream` as a function in the linter config as is done for `cmd`? Seems like that could work if backwards compatibility is necessary.

cpof-tea:
> Otherwise the linter could be implemented as `function()`

I didn't know it is supported, that's why I asked.

> first calls `stylelint --version` (if that's supported), parses the version and depending on that returns a different table. But not sure if that's worth it.

Why not. At least it feels more natural, than just pretending we know if other streams used or not.

I will try to implement

mfussenegger:
Could you first try setting the stream to `both`? If that works, it's a much simpler solution.

cpof-tea:
@mfussenegger okay, it works.
@jmpolom could you please set `both` to `stream` option?

albnavarro:
I think this pr can be merged as is `stream = "sterr" ` so that it is compatible with the stable version of stylelint.
Personally, to have backward compatibility I inserted:

`
 local stylelint = require("lint").linters.stylelint
`
`
stylelint.stream = "both"
`

It may be correct that the choice to use `both` is optional.

jmpolom:
It seemed that the change by Stylelint to output errors to `stderr` was permanent. Based on that I agree with @albnavarro and this should be merged as-is. This module should support the current stable upstream linter version which outputs lint messages to `stderr`.

mfussenegger:
https://github.com/mfussenegger/nvim-lint/pull/592

PR #529, opened 2024-02-07
fix(typos): line_num is not always returned

WhoIsSethDaniel:
At least with recent `typos` (I have 1.18.1) the `line_num` is not returned when the typo is discovered on the first line of the document. Only `byte_offset` is returned. This causes a *parser error* to appear at the top of the file in vtext.

### DIFF
--- a/lua/lint/linters/typos.lua
+++ b/lua/lint/linters/typos.lua
@@ -28,7 +28,7 @@ return {
     for json in string.gmatch(output, "[%S]+") do
       local item = vim.json.decode(json)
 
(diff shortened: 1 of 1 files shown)


WhoIsSethDaniel:
I have tested back to 1.16.1 and this behavior is consistent to at least that release.

WhoIsSethDaniel:
This only happens on *filenames* that are misspelled. I have chosen to ignore the error with this PR.

mfussenegger:
Thanks

PR #530, opened 2024-02-07
feat: add spectral

luispflamminger:
Hi @mfussenegger,

thanks for this great plugin.  
I would like to add the [Spectral](https://docs.stoplight.io/docs/spectral/674b27b261c3c-overview) linter for OpenAPI specifications.

There is a slight problem in that the linter takes a *required* parameter `--ruleset` that takes either a file path or a URL.  
Without this parameter, the linter will error out.

I have implemented it in a way, where the user would have to do something like this, which is a bit awkward 😄
```lua
require('lint').linters.spectral = require('lint').linters.spectral("path/to/ruleset")
```

Another approach would be to not specify the ruleset parameter by default, let the linter error and let the user figure out how to override the linters `args`. This is also pretty inconvenient.

I would appreciate your feedback on which of the options you would prefer, or if you maybe have a better idea.  
I could also contribute some better documentation so users would know how to configure this specific linter.

If it's not a good fit for the project because of this constraint, I obviously understand.

### DIFF
--- a/lua/lint/linters/spectral.lua
+++ b/lua/lint/linters/spectral.lua
@@ -1,0 +1,50 @@
+local severities = {
+  vim.diagnostic.severity.HINT,
+  vim.diagnostic.severity.INFO,
+  vim.diagnostic.severity.WARN,
+  vim.diagnostic.severity.ERROR,
+}
+
+return function(ruleset)
+  return {
+    cmd = 'spectral',
+    stdin = false,
+    append_fname = true,
+    args = { "lint", "-r", ruleset, "-f", "json", },
+    stream = "stdout",
+    ignore_exitcode = true,
+    parser = function(output, _)
+      local items = {}
+
+      if output == '' then
+        return items
+      end
+
+      local decoded = vim.json.decode(output) or {}
+      local bufpath = vim.fn.expand('%:p')
+
+      -- prevent warning on files that are not OpenAPI specs
+      if decoded[1].code == "unrecognized-format" then
+        return items
+      end
+
+      for _, diag in ipairs(decoded) do
+        vim.print(diag.severity)
+        if diag.source == bufpath then
+          table.insert(items, {
+            source = "spectral",
+            severity = severities[diag.severity + 1],
+            code = diag.code,
+            message = diag.message,
+            lnum = diag.range.start.line + 1,
+            end_lnum = diag.range["end"].line + 1,
+            col = diag.range.start.character + 1,
+            end_col = diag.range["end"].character + 1,
+          })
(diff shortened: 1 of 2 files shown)


SirAppSec:
Thank you so much, I was about to create a PR for this.

A few checks before merging this PR.
1. Would extending Spectral with npm packages work put of the box?
2. Should a user install Spectral dependencies for this to work?

luispflamminger:
Thanks for the feedback @SirAppSec!

I don't think having specific logic for dependency issues is in scope for this plugin. We simply want to execute the linter as it is configured on the users system.

I will build some basic error handling when I find the time, because right now there is no way of knowing whether the linter errored, which is pretty confusing. I would also like to throw an error when spectral is not in PATH.

Your point about `.spectral.yaml` is good, I didn't think about that. I will change it as suggested.
Users will then have the option to override the ruleset using custom `args`.

luispflamminger:
I removed the `--ruleset` option by default. Spectral will now look for a `.spectral.*` file by default, as specified [in the docs](https://docs.stoplight.io/docs/spectral/9ffa04e052cc1-spectral-cli#using-a-ruleset-file). This also means users can now use this like they would any other linter in this plugin.

I also improved error handling so users get notified if no ruleset was found (`WARN`) or if Spectral didn't return valid JSON (`ERROR`), which happens when Spectral itself ran into an error. You can test this by specifying an invalid `.spectral.yaml`, for example.

The case that Spectral can't be found on PATH is already handled by the plugin itself, so nothing to do here.

@SirAppSec: Please feel free to take another look and give feedback 🙂

lua/lint/linters/spectral.lua
```diff
@@ -0,0 +1,56 @@
...
+      vim.notify(no_ruleset_found_msg, vim.log.levels.WARN)
+      return {}
+    end
+
+    -- attempt to decode linter result
+    local success, result = pcall(vim.json.decode, output)
+    if not success then
+      vim.notify("Spectral exited with an error:\n" .. output, vim.log.levels.ERROR)
+      return {}
+    end
```
mfussenegger, line 30:
```suggestion
    result = vim.json.decode(output)
```

You can just let it fail, there is default error handling: https://github.com/mfussenegger/nvim-lint/blob/03b1fc593638098a35de26d768d5f43b0fe57041/lua/lint/parser.lua#L143-L155

mfussenegger:
Could you address the suggestion?

luispflamminger:
@mfussenegger Sorry it took so long. I addressed your suggestion and rebased the branch.

PR #533, opened 2024-02-09
feature: add CUE linter

mhersson:
adding new linter for the CUE language using the `cue vet` command

### DIFF
--- a/lua/lint/linters/cue.lua
+++ b/lua/lint/linters/cue.lua
@@ -1,0 +1,12 @@
+local efm = "%E%m:,%C%f:%l:%c"
+
+return {
+  cmd = "cue",
+  args = { "vet" },
+  stdin = false,
+  ignore_exitcode = true,
+  stream = "stderr",
+  parser = require("lint.parser").from_errorformat(efm, {
+    source = "cue vet",
+  }),
+}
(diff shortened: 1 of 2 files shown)


lua/lint/linters/cue.lua
```diff
@@ -0,0 +1,38 @@
...
+    local diagnostics = {}
+    local lnum, col, message
+    local pattern = ":(%d+):(%d+)$"
+
+    for _, line in ipairs(vim.fn.split(output, "\n")) do
+      if line:match("^%s+") then
+        lnum, col = line:match(pattern)
+      else
+        message = line:gsub(":", "")
+      end
```
mfussenegger, line 22: Could the parser use `from_errorformat` or `from_pattern` instead - looks like it can be declared with a pattern?

mhersson: Can I? 
The output from `cue vet` spans over 2 lines where the message is on the first line and then the location follows on the next line.
Like so:
```
❯ cue vet example.cue
square bracket must have exactly one element:
    ./example.cue:19:14
string literal not terminated:
    ./example.cue:25:34
```
The readme says 
> The regular expression pattern applied on each line of the output

mhersson:
Pushed new version using  `from_errorformat`

PR #537, opened 2024-02-16
sqlfluff: change default dialect to ansi

### LINKED ISSUE #441: The default SQL dialect in SQLFluff
https://github.com/mfussenegger/nvim-lint/blob/962a76877a4479a535b935bd7ef35ad41ba308b2/lua/lint/linters/sqlfluff.lua#L3-L7

The default SQL dialect  should change from `postgres` to `ansi`

> ANSI
> This is the base dialect which holds most of the definitions of common SQL commands and structures. If the dialect which you’re actually using isn’t specifically implemented by SQLFluff, using this dialect is a good place to start.
[Doc of SQLfluff](https://docs.sqlfluff.com/en/stable/dialects.html#ansi)

fbuchlak:
Resolves #441

### DIFF
--- a/lua/lint/linters/sqlfluff.lua
+++ b/lua/lint/linters/sqlfluff.lua
@@ -3,7 +3,7 @@ return {
   args = {
     "lint", "--format=json",
     -- note: users will have to replace the --dialect argument accordingly
-    "--dialect=postgres",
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks

PR #548, opened 2024-03-01
new function -> get_running_count()

daUnknownCoder:
Checks number of current linters

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -196,6 +196,27 @@ end
 ---@type table<integer, table<string, lint.LintProc>> bufnr: {linter: handle}
 local running_procs_by_buf = {}
 
+--- Returns the number of the running linters
+---
+---@param bufnr? integer buffer for which to get the running linters. nil=all buffers
+---@return string[]
+function M.get_running_count(bufnr)
+  local count = 0
+  if bufnr then
+    bufnr = bufnr == 0 and api.nvim_get_current_buf() or bufnr
+    local running_procs = (running_procs_by_buf[bufnr] or {})
+    for _, _ in pairs(running_procs) do
+      count = count + 1
+    end
+  else
+    for _, running_procs in pairs(running_procs_by_buf) do
+      for _ in pairs(running_procs) do
+        count = count + 1
+      end
+    end
+  end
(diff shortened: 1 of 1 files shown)


daUnknownCoder:
> Thanks, but you can just use `#require("lint").get_running()`.

i was struggling to use this:
```lua
          {
            function()
              local lint_progress = function()
                local linters = require("lint").get_running()
                if #linters == 0 then
                  return "󰦕 Linting completed"
                end
                return "󱉶 Linting with '" .. table.concat(linters, ", ") .. "' in progress"
              end
              return lint_progress()
            end,
          },
```

i m using `lualine` but, it gives me `<icon> Linting completed` for a non linted file too, like if i do `nvim ifregre.reger` it still gives me linting completed but there is no linting happening... which is why i was fiddling with that functions... can you help me with that?

mfussenegger:
You can specify the buffer number for which to get the running linters and use 0 for current buffer - e.g.: `require("lint").get_running(0)`

daUnknownCoder:
> `require("lint").get_running(0)`

I mean that when I use get_running, if the current buffer's filetype has no linter installed, there should be `no linter found` instead of `linting completed` as per `#linters==0`, I want to have a get a `if elseif else`, like
```lua
if #linters==none
     return "No linters found"
elseif #linters==0
     return "Linting completed"
else #linters>0
     return "Linting in progress"
```

Can I get something like that?

mfussenegger:
The problem with that is that users can call `try_lint` _without_ having configured the linter.
nvim-lint doesn't have the complete information over which linters could be triggered and how.

Because of that it's up to users to decide how & what information to display exactly.
`get_running` only displays the number of _running_ linters, not _configured_ linters. Having a `get_running_linter_count()` method that returns the `_configured_ linters would be misleading, and could be incorrect. For example I have a `require("lint").try_lint("ruff")` call in a `BufWritePost` autocmd, and `ruff` is not in the `linters_by_ft` table

daUnknownCoder:
so from what i understand, there's no way to achieve `no linters found` like i just wanna check the `ft` and if it does not have a linter or if the linter is not installed

Issue #355, opened 2023-08-30
How to use same tool in mutiple filetypes


caiyanhu:
I want use cspell in some filetypes.
This is my configure:
			require("lint").linters_by_ft = {
				css = { "stylelint", "cspell" },
				javascript = { "eslint", "cspell" },
				javascriptreact = { "stylelint", "eslint", "cspell" },
				typescript = { "eslint", "cspell" },
				typescriptreact = { "stylelint", "eslint", "cspell" },
				-- html = { "cspell" },
				-- json = { "cspell" },
				-- lua = { "cspell" },
				-- rust = { "cspell" },
				-- vue = { "cspell" },
				["*"] = { "cspell" },

cspell not work when set on wildcard, any way to do so? thx!

caiyanhu:
this snippet works for me
```lua
			local languages = {
				"css",
				"html",
				"javascript",
				"javascriptreact",
				"json",
				"lua",
				"rust",
				"typescript",
				"typescriptreact",
				"vue",
			}
			for _, name in pairs(languages) do
				if not linters[name] then
					linters[name] = { "cspell" }
				end
				if not vim.tbl_contains(linters[name], "cspell") then
					table.insert(linters[name], "cspell")
				end
			end

```

chrisgrieser:
yeah, I had the same issue (only with `codespell`). Having to use a loop like this is quite inconvenient, and also reduces readability.

formatter.nvim and conform.nvim both have the "special" filetype `["*"]` which adds the tool to all filetypes. Maybe nvim-lint could also support that?

mfussenegger:
You can add a extra `try_lint("cspell")` call. For example:

```lua
local lint = require("lint")
lint.linters_by_ft = {
  css = { "stylelint" },
  javascript = { "eslint" },
  javascriptreact = { "stylelint", "eslint" },
  typescript = { "eslint" },
  typescriptreact = { "stylelint", "eslint" },
}

vim.api.nvim_create_autocmd({"BufWritePost", "BufEnter"}, {
  group = vim.api.nvim_create_augroup('lint', { clear = true }),
  callback = function()
    lint.try_lint()
    lint.try_lint("cspell")
  end
end,
}
```

DownRangeDevOps:
> You can add a extra `try_lint("cspell")` call. For example:
> 
> ```lua
> local lint = require("lint")
> lint.linters_by_ft = {
>   css = { "stylelint" },
>   javascript = { "eslint" },
[...]

PR #549, opened 2024-03-04
feat: add configuration of global (`*`) and fallback (`_`) linters

mehalter:
This adds the ability to provide both global and fallback linters which play nicely with the filetype resolution. I also added a couple tests to make sure they get applied correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,10 @@ git clone \
 Configure the linters you want to run per file type. For example:
 
 ```lua
-require('lint').linters_by_ft = {
-  markdown = {'vale',}
+require("lint").linters_by_ft = {
+  markdown = { "vale" },
+  ["*"] = { "codespell" }, -- global filetype to run linter on all filetypes
+  ["_"] = { "write-good" }, -- fallback filetype to run linter when no filetype specific linter is available
 }
 ```
 
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -79,22 +79,25 @@ local function read_output(cwd, bufnr, parser, publish_fn)
   end
 end
 
-
-function M._resolve_linter_by_ft(ft)
-  local names = M.linters_by_ft[ft]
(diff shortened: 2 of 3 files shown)


mfussenegger:
Thanks but the way to add fallback linter is to add additional `try_lint` calls: https://github.com/mfussenegger/nvim-lint/issues/355#issuecomment-1759203127

mehalter:
Okie dokie. This is how I currently do it as well. I noticed that this code seems to be duplicated basically in most configurations I see that use `nvim-lint` which seems like it would be a good candidate to include upstream. Also because it would enable much better support for using the `_resolve_linter_by_ft` function for building statusline entities rathern than using `get_running` since that is not the most useful if I want to see a list of linters that are configured to run in the buffer. Totally understand if you don't want to support this out of the box though!

fresh2dev:
Found this out PR due to my own confusion, assuming that `nvim-lint` could be configured like `conform.vim`.

This PR enables specifying both *global* linters and/or a *fallback* linters. The workaround in #355 only makes it possible to specify *global* linters.

mehalter:
@fresh2dev you can leverage the fact that Lua is pretty loose and do this by overriding the internal Lua API in `nvim-lint`. Here is an example that is similar to how I'm doing this now in my personal configuration:

```lua
local lint = require "lint"

local orig_resolve_linter_by_ft = lint._resolve_linter_by_ft
lint._resolve_linter_by_ft = function(...)
  local linters = orig_resolve_linter_by_ft(...)
  if not next(linters) then linters = lint.linters_by_ft["_"] end -- add fallback linters
  vim.list_extend(linters, lint.linters_by_ft["*"]) -- add global linters

  return linters
end
```

Hopefully there are no breaking changes to this `_resolve_linter_by_ft` function 😅

fresh2dev:
Awesome; thank you!

PR #550, opened 2024-03-07
feat: implement temp file

### LINKED ISSUE #235: Feature request: shadow file / temporary file as when `stdin = false`
Hi, there

Great plugin you got going. I like it that it's easy to configure and/or extend since it's written in lua.

One thing I'm missing is being able to use a temporary file when `stdin = false` (namely, with `mypy`). This way I don't _need_ to save my file to get lint the current buffer.

I think I can implement this if maintainers are open to dealing with ~~opening pandora's box~~ this rather complicated featuer. My belief is that, while it may lead to lots of corner cases, these can and will be covered through usage within few months (ie: this is a feature that'll take some time to stabilize)

see https://github.com/neomake/neomake/commit/4bb03bf7343357f241b763018a078ed6f3d163e1

FelipeLema:
fixes #235

### DIFF
--- a/lua/lint/temp_file.lua
+++ b/lua/lint/temp_file.lua
@@ -1,0 +1,55 @@
+local M = {}
+
+---@class lint.temp_file.Opts
+---@inlinedoc
+---
+--- Extension for the filepath
+---@field ext? string
+---
+--- Timeout in ms to delete the file (defaults to 30_000 = 30s)
+---@field timeout? number
+---
+--- Text to be inserted into file
+---@field str? string
+
+--- Get a temporary filepath. This filepath is deleted on nvim exit or after timeout
+---@param opts? table
+---@return string filepath to temp file
+function M.temp_filepath(opts)
+  opts = opts or {}
+  local temp_filepath = assert(os.tmpname())
+  if opts.ext then
+    temp_filepath = temp_filepath .. '.' .. opts.ext
+  end
+
+  local timer_r, vimexit_r
+
+  timer_r = vim.defer_fn(function()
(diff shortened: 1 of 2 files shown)


FelipeLema:
missing tests

FelipeLema:
PR ready, although I'm unsure why 0.6.1 is failing

FelipeLema:
oh, so `nvim_create_autocmd` did not exist ... not sure if it's worth re-doing using vimscript, since it's a (lua / luv) timer object that needs to be handled in there

dante0624:
@FelipeLema I'm new to this plugin and don't fully understand it, but right now it looks like your changes are unused. Before merging this, wouldn't some type of modification be needed to the init.lua file's `M.lint` function? It would need to call your `temp_filepath` function, actually create the temp file, fill it with the contents of the current buffer, and then append the filename to the `args` table (in `M.lint`)?

Also, the creation of the temp file and filling it with the contents of the current buffer could all happen within your `temp_file.lua` file. Maybe a new function or two could be created for that.

I would be happy to work on this with you by the way. Let me know what you think

FelipeLema:
hey @dante0624 

the scope of this PR is within https://github.com/mfussenegger/nvim-lint/issues/235 and is meant to provide a tool for shadw files when needed. it is not my intention to replace default implementation of this plugin, but rather extend it to i cover previously ignored usecases

filling the tempfile could be done in different ways. I explicitly left that out to keep this PR short

mfussenegger:
Thanks for the PR but I imagine tempfile support to work a bit differently. Not going to merge this.

PR #554, opened 2024-03-10
feat(parser): adds option to not split lines

AckslD:
I don't know if this is the best solution but I was having issues with the `nix` linter. The reason seems to be that I get errors such as:
```
error: undefined variable 'pkg'
       at /home/axel/.config/home-manager/home.nix:23:5:
           22|     # # "Hello, world!" when run.
           23|     pkg.hello
             |     ^
           24|     pkgs.neovim
```
where the pattern (`(%w+): (.+)\n%s*at .+:(%d+):(%d+)`) looks for a match which is actually in the above two lines. Since lines from stdout/stderr are passed line by line to the pattern we don't get any match.

This adds an option to `from_pattern` to not split the lines before passing to the pattern. This can of course then only give a single error at a time. But at least for `nix` this seems to be the case anyway?

Happy for some suggestions of how to handle this better.

### DIFF
--- a/lua/lint/linters/nix.lua
+++ b/lua/lint/linters/nix.lua
@@ -1,4 +1,4 @@
-local pattern = '^(%w+): (.+) at .+:(%d+):(%d+)$'
+local pattern = '(%w+): (.+)\n%s*at .+:(%d+):(%d+)'
 local groups = { 'severity', 'message', 'lnum', 'col' }
 local severity_map = { error = vim.diagnostic.severity.ERROR }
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
Does nix support different output formats? 

Otherwise errorformat would support multi-line handling and there's the option to implement a `function(output, bufnr)` directly, which I think I'd prefer over adding a new option to `from_pattern` that limits matching to a single diagnostic result.

Issue #569, opened 2024-04-13
error message every time i open or save a file which linter is not installed


daniele821:
nvim-lint plugin report the following error message:

```
Error detected while processing BufEnter Autocommands for "*":
Error running markdownlint: ENOENT: no such file or directory
```
every time i open or save a file which has a linter associated, but the linter is not installed.
For example i declare "shellcheck" as the linter for bash, but i do not install shellcheck, and then i try to open a bash file, the error message written above will show up once i open the file, and every time i save the file.

I would like if it was like conform (which takes care of formatting), where if you don't have the thing installed, it just doesn't run.

thanks

ivanesmantovich:
Hi!
`try_lint` has an `ignore_errors` option to ignore command-not-found errors: `require("lint").try_lint(nil, { ignore_errors = true })`
#454

Issue #454, opened 2023-11-10
Eslint error when .eslint.json doesn't exist.


matthewrosse:
Hello, I get an error `Error detected while processing BufWritePost Autocommands for "*":
Error running eslint: ENOENT: no such file or directory`. Is there a way not to trigger linter if it cannot find its configuration file? Because in some quick and dirty projects I don't setup eslint and its annoying when it shows the error on every save.

caiyanhu:
I use lerna to manage projects sometimes, if I have installed eslint for example in root, the inner project will still report the above error. Seems like it won't look up eslint in upper space.

mfussenegger:
`try_lint` has an `ignore_errors` option to ignore command-not-found errors: `require("lint").try_lint(nil, { ignore_errors = true })`

PR #570, opened 2024-04-14
Add ignore_errors option to the README

ivanesmantovich:
People keep asking about the option to ignore ENOENT errors (#454, #569), i was also having the same problem
I think it's a good idea to include it in the README!


mfussenegger:
Thanks but my plan is to eventually generate API docs under a `doc/lint.txt`

PR #586, opened 2024-05-23
fix(eslint): warn user of no config file instead of showing in buffer

### LINKED ISSUE #462: Eslint_d error when .eslint.json doesn't exist.
I have the the same issue as mentioned in https://github.com/mfussenegger/nvim-lint/issues/454#issue-1986742975 except for with **eslint_d** and solution mentioned there wouldn't work, so setting `ignore_errors = true` option for `try_lint` wouldn't work. https://github.com/mfussenegger/nvim-lint/blob/9e096df0858a1834ec2a60abe6cceb4d84dbd34c/lua/lint.lua#L208 seems to return as successful, but in the buffer I see the error:
```
Could not parse linter output due to: Expected value but found invalid token at character 1
output: Error: No Eslint configuration found in <path-to-project-root>.
```
I'm not sure how to go on debugging this issue.

It works as expected with eslint.

IamGideonIdoko:
Closed #462

When there's no ESLint config, the error is shown in the buffer, which is an annoying experience when a user is working on JS-related files that are not associated with a config.

Screenshot 2024-05-23 at 1 56 57 PM

We should warn the user of the error once instead of showing the error in the buffer:

Screenshot 2024-05-23 at 2 10 09 PM

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -20,11 +20,16 @@ return {
   stream = 'stdout',
   ignore_exitcode = true,
   parser = function(output, bufnr)
-    if vim.trim(output) == "" then
+    local trimmed_output = vim.trim(output)
+    if trimmed_output == "" then
       return {}
     end
     local decode_opts = { luanil = { object = true, array = true } }
     local ok, data = pcall(vim.json.decode, output, decode_opts)
+    if string.find(trimmed_output, "No ESLint configuration found") then
+      vim.notify_once(trimmed_output, vim.log.levels.WARN)
+      return {}
+    end
     if not ok then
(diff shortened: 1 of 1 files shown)


ankitdev10:
Legit fix. Not having eslint config file in the directory had been troubling me for quite a long.

IamGideonIdoko:
@mfussenegger What do you make of this?

Prashant20nov2003:
please fix this issue its super annoying.

AlejandroSanchez90:
I'm still experiencing this, no eslint config, and seems
1. Could not parse linter output due to: Expected value but found invalid token at character 1
   output: Error: Could not find config file.

PR #589, opened 2024-05-24
trivy: bugfix: line reported as col; bugfix: crash on missing line from trivy; plus refinements

bobsrentacow:
**Changes**

* bugfix: no longer reports trivy StartLine, EndLine as nvim-lint col, end_col
* bugfix: check parsed fields before building/inserting diagnostic
* cleanup: fix deprecated CLI arg: '--scanners config' -> '--scanners misconfig'
* comments: clarify intent to ease future maintainence

**Bug Descriptions**

Output from trivy sometimes includes line info, but not column info.  line info is used for both col and lnum in trivy.lua

Output from trivy does not always contain StartLine and EndLine fields.  When they are missing, diagnostics are not created for each Misconfiguration from trivy, but instead a single diagnostic is generated with a lua error about accessing a nil value (StartLine).

To reproduce, open the following Dockerfile with trivy enabled as a linter, then look at the reported diagnostics:
```
FROM ubuntu:24.04
LABEL maintainer="John Doe <john.doe@aol.com>"
```

### DIFF
--- a/lua/lint/linters/trivy.lua
+++ b/lua/lint/linters/trivy.lua
@@ -9,11 +9,10 @@ return {
   cmd = "trivy",
   stdin = false,
   append_fname = true,
-  -- FIXME: '--scanners config' is deprecated in favor of '--scanners misconfig'
   -- NOTE: excluding '-q/--quiet' flag to receive log on stderr before the JSON on stdout.
   --       The log may contain errors and warnings that can be used to notify the user this script
   --       needs maintenance.
-  args = { "--scanners", "config", "--format", "json", "fs" },
+  args = { "--scanners", "misconfig", "--format", "json", "fs" },
   stream = "stdout",
   ignore_exitcode = false,
   parser = function(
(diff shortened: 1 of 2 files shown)


lua/lint/linters/trivy.lua
```diff
@@ -8,24 +9,59 @@ return {
...
-  parser = function(output, bufnr)
+  parser = function(
+    output, -- output from trivy (a JSON string)
+    bufnr -- nvim buffer number. Used to choose the correct Result from the JSON output
+  )
     local diagnostics = {}
+
+    -- TODO: Parse log messages from trivy that appear before the JSON.  Register any ERRORS or
+    --       WARNINGS as diagnostics so someone will know to make future updates to this script.
+    -- NOTE: Don't know how to get access to both stderr and stdout, so unable to address this
```
bobsrentacow, line 26: HELP: What is mechanism to report errors/warnings during linter run?  That handled elsewhere?

PR #590, opened 2024-05-27
Read stylelint output from stderr

sqwxl:
The latest major release of `stylelint`, v16, [changed](https://stylelint.io/migration-guide/to-16/#changed-cli-to-print-problems-to-stderr) the program to output problems to stderr rather than to stdout, as before.

I have tested the proposed changed locally and have found it to work as expected.

### DIFF
--- a/lua/lint/linters/stylelint.lua
+++ b/lua/lint/linters/stylelint.lua
@@ -22,7 +22,7 @@ return {
       return vim.fn.expand("%:p")
     end,
   },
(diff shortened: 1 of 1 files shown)


sqwxl:
Closing this as a duplicate of #590

mfussenegger:
As mentioned in the other PR I think using `stream = 'both'` should work for both versions and I'd merge a PR that makes that change

PR #598, opened 2024-06-07
feat: nilaway linter support

NormanPerrin:
Thanks for the plugin!

This adds support for the [nilaway](https://github.com/uber-go/nilaway) linter. Not much to add

### DIFF
--- a/lua/lint/linters/nilaway.lua
+++ b/lua/lint/linters/nilaway.lua
@@ -1,0 +1,55 @@
+return {
+  cmd = 'nilaway',
+  stdin = false,
+  append_fname = true,
+  args = { '-pretty-print=false', '-json' },
+  stream = 'stdout',
+  ignore_exitcode = true,
+  parser = function(output)
+    local diagnostics = {}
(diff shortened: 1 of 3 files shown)


lua/lint/linters/nilaway.lua
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      table.insert(diagnostics, {
+        lnum = l - 1,
+        col = c - 1,
+        message = diagnostic.message,
+        severity = vim.diagnostic.severity.WARN
+      })
+
+      ::continue::
```
mfussenegger, line 50:
```suggestion
      if l and c then
        table.insert(diagnostics, {
            lnum = l - 1,
            col = c - 1,
            message = diagnostic.message,
            severity = vim.diagnostic.severity.WARN
        })
      end
```

`goto` is not available in puc lua 5.1, only with luajit and neovim supports both

NormanPerrin:
Closing for now as this only works on a per-file basis

PR #599, opened 2024-06-08
Add VSG linter

fredrikfoss:
Adds the [VHDL Style Guide](https://github.com/jeremiah-c-leary/vhdl-style-guide) (VSG) linter.

The linter doesn't autodetect configuration files, so I added functionality to look for `.vsg.yaml` and `.vsg.json` in the file's directory and in `xdg-config/vsg` or `~/.config/vsg`. I could not find any mention of a standard name for said files in the documentation, but I read some people using these names and so do I. I'm not sure if this is out of scope, but it is quite nice. Please let me know.

### DIFF
--- a/lua/lint/linters/vsg.lua
+++ b/lua/lint/linters/vsg.lua
@@ -1,0 +1,45 @@
+local pattern = "(%w+).*%((%d+)%)(.*)%s+%-%-%s+(.+)"
+local groups = { "severity", "lnum", "code", "message" }
+local severity_map = {
+  ["ERROR"] = vim.diagnostic.severity.ERROR,
+  ["WARNING"] = vim.diagnostic.severity.WARN,
+  ["INFORMATION"] = vim.diagnostic.severity.INFO,
+  ["HINT"] = vim.diagnostic.severity.HINT,
+}
+
+local function find_local_config()
+  local current_file = vim.api.nvim_buf_get_name(0)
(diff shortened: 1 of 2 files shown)


fredrikfoss:
Hi, added some changes to better match with conform.nvim, see [#451](https://github.com/stevearc/conform.nvim/pull/451). Also resolved a bug conserning the discovery of config files.

I'm not very experienced with Lua, but could someone point out why this didn't work:

```lua
return {
    cmd = "vsg",
    stdin = true,
    args = function()
        local args = { "-of", "syntastic", "--stdin" }
        local config_file = find_config(vim.fn.expand("%:p:h"))

        if config_file then
            table.insert(args, "-c")
            table.insert(args, config_file)
        end

        return args
    end,
    stream = "stdout",
    ignore_exitcode = true,
    parser = require("lint.parser").from_pattern(pattern, groups, severity_map, {
        source = "vsg",
    }),
}
```

lua/lint/linters/vsg.lua
```diff
@@ -0,0 +1,63 @@
...
+    table.insert(args, config_file)
+  end
+
+  return args
+end
+
+return {
+  cmd = "vsg",
+  stdin = true,
+  args = get_args(vim.fn.expand("%:p:h")),
```
mfussenegger, line 57:
```suggestion
  args = function()
    return get_args(vim.fn.expand("%:p:h"))
  end,
```

Otherwise this gets executed once when the module is first loaded and is then frozen.

PR #605, opened 2024-06-23
Ppass cwd to linter command

thunderer199:
Node.js linters (eslint, stylelint, etc) run binary from relative `node_modules` that depend on the working directory. But not always working directory has `node_modules`, in the case of monorepo it can be some sub-folder.

The lint run command already has `cwd` option, but doesn't pass it to `cmd` command.
The fix is passing opts to linter `cmd` to properly calculate the path to `node_modules` if it's not in working directory.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -264,9 +264,9 @@ function M.try_lint(names, opts)
   running_procs_by_buf[bufnr] = running_procs
 end
 
-local function eval_fn_or_id(x)
+local function eval_fn_or_id(x, opts)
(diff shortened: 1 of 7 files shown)


fredrikaverpil:
Hey @thunderer199 FYI, please note I opened another PR which touches `cwd` in #673

mfussenegger:
@thunderer199 thanks for the PR - could you try https://github.com/mfussenegger/nvim-lint/pull/735 instead?

I'd like to avoid that all linter implementations have to deal with the `cwd` option.

PR #616, opened 2024-07-08
feat(lint): use function as args

xudyang1:
Currently, `require('lint').linters.LINTER_NAME.args` only supports `(string|fun():string)[]` type. Although it supports entries of `fun():string` type, inserting multiple options in the table is still cumbersome.

Changing the args to `fun(bufnr: integer): string[]` provides more fine-grained control. However, this introduces a breaking change to current config.

This pr sets old args type as `lint.Args` or `(string|fun():string)[]`, and uses `lint.Args | fun(bufnr:integer): lint.Args` as the new type. This enables use of a function as args without breaking changes.

This pr makes the following changes to linters possible:

```lua
---Find pattern path from buf dir to $HOME
---@param pattern string file or directory name pattern
---@return string | nil full path of the pattern if found, nil otherwise
local function find_path(pattern)
  return vim.fs.find(pattern, {
    upward = true,
    -- stops at $HOME/.., so search includes $HOME
    stop = vim.fn.expand("~/.."),
    path = vim.fs.dirname(vim.api.nvim_buf_get_name(0)),
  })[1]
end

local lint = require('lint')
lint.linters.selene.args = function()
  local selene_config_option = "--config"
  local selene_toml = find_path("selene.toml")
  if selene_toml then
    return { "--display-style", "json", selene_config_option, selene_toml, "-" }
  else
    return { "--display-style", "json", "-" }
  end
end

lint.linters.luacheck.args = function()
  local luacheckrc_config_option = "--config"
  local luacheckrc = find_path(".luacheckrc")
  if luacheckrc then
    return { "--formatter", "plain", "--codes", "--ranges", luacheckrc_config_option, luacheckrc, "-" }
  else
    return { "--formatter", "plain", "--codes", "--ranges", "-" }
  end
end
```

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -8,11 +8,12 @@ local M = {}
 ---@field on_chunk fun(chunk: string)
 ---@field on_done fun(publish: fun(diagnostics: vim.Diagnostic[]), bufnr: number, linter_cwd: string)
 
+---@alias lint.Args (string|fun():string)[]
 
 ---@class lint.Linter
 ---@field name string
 ---@field cmd string
----@field args? (string|fun():string)[] command arguments
+---@field args? lint.Args | fun(bufnr: integer): lint.Args command arguments
 ---@field stdin? boolean send content via stdin. Defaults to false
 ---@field append_fname? boolean add current file name to the commands arguments
 ---@field stream? "stdout"|"stderr"|"both" result stream. Defaults to stdout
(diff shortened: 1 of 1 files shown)


mfussenegger:
Any reason returning the whole linter as a function is not sufficient? See the `compiler` linter for an example

xudyang1:
That works, but I have to make a deepcopy of the original linter.

```lua
local old_selene = vim.deepcopy(lint.linters.selene)
lint.linters.selene = function()
  local selene_config_option = "--config"
  local selene_toml = find_path("selene.toml")
  if selene_toml then
    return vim.tbl_extend(
      "force",
      old_selene,
      { args = { "--display-style", "json", selene_config_option, selene_toml, "-" } }
    )
  else
    return old_selene
  end
end
```

xudyang1:
> Any reason returning the whole linter as a function is not sufficient? See the `compiler` linter for an example

That works, but I have to make a deepcopy of the original linter.

```lua
local old_selene = vim.deepcopy(lint.linters.selene)
lint.linters.selene = function()
  local selene_config_option = "--config"
  local selene_toml = find_path("selene.toml")
  if selene_toml then
    return vim.tbl_extend(
      "force",
      old_selene,
      { args = { "--display-style", "json", selene_config_option, selene_toml, "-" } }
    )
  else
    return old_selene
  end
end
```

xudyang1:
Setting `cwd` in `lint.linters.selene` also achieves the same goal. However, `cwd` only supports `string` type, and thus we have to maintain a deepcopy too.

```lua
local old_selene = vim.deepcopy(lint.linters.selene)
lint.linters.selene = function()
  local cwd = vim.fs.root(0, { ".git", "selene.toml" })
  old_selene.cwd = cwd
  return old_selene
end
```

If `cwd` supports `fun(): string`, then we don't have to maintain the copies in our configuration, and can simply updates the `cwd` by:

```lua
lint.SOME_LINTER.cwd = function() return vim.fs.root(0, SOME_LINTRC_FILE) end
```

PR #621, opened 2024-07-19
feat: add eugene migration linter for postgres

WhoIsSethDaniel:


lua/lint/linters/eugene.lua
```diff
@@ -0,0 +1,31 @@
...
+    '--format=json',
+  },
+  stdin = false,
+  parser = function(output, _)
+    local diagnostics = {}
+    if #output > 0 then
+      local ok, decoded = pcall(vim.json.decode, output)
+      if not ok then
+        return diagnostics
+      end
```
mfussenegger, line 14:
```suggestion
      local decoded = vim.json.decode(output)
```

Decode errors would be caught further up in the framework and the error gets reported as diagnostic in the first line.

That can help catch configuration errors.

lua/lint/linters/eugene.lua
```diff
@@ -0,0 +1,31 @@
...
+              source = 'eugene',
+              lnum = stmt.line_number,
+              col = 0,
+              severity = vim.diagnostic.severity.ERROR,
+              message = string.format('%s: %s: %s', tr.id, tr.name, tr.url),
+            })
+          end
+        end
+      end
+    end
```
mfussenegger, line 28:
```suggestion
        for _, tr in ipairs(stmt.triggered_rules) do
          table.insert(diagnostics, {
            source = 'eugene',
            lnum = stmt.line_number,
            col = 0,
            severity = vim.diagnostic.severity.ERROR,
            message = string.format('%s: %s: %s', tr.id, tr.name, tr.url),
          })
        end
      end
    end
```

The length check is kinda redundant

mfussenegger (commented):

WhoIsSethDaniel:
Added to README.
Applied suggested changes.

PR #623, opened 2024-07-24
fix(credo): credo now respects .credo.exs

georgeguimaraes:
I changed credo spec to pass around the filename. This is important because credo has defaults that disable some checks for certain directories. And if the repo has a `.credo.exs` file, credo needs to know the filename that it is checking.

I also created the severity mapping.

(I could not run the tests on my machine, so I want to run the CI to see if everything works).


georgeguimaraes:
Got busted working.
Tests are fixed, should be ok now.

mfussenegger:
cc @skbolton does this look okay to you?

PR #626, opened 2024-07-29
feat: add clippy

wispl:
Rust linting using cargo clippy, done to the best of my limited abilities. Since clippy lints the whole project, some filtering has to be done. I am not too keen on the recursion but sometimes there is useful information from the child messages. Thanks!

### DIFF
--- a/lua/lint/linters/clippy.lua
+++ b/lua/lint/linters/clippy.lua
@@ -9,7 +9,7 @@ local function parse(diagnostics, file_name, item)
     if span.file_name == file_name then
       local message = item.message
       if span.suggested_replacement ~= vim.NIL then
-        message = message .. " (" .. tostring(span.suggested_replacement) .. ")"
+        message = message .. "\nSuggested replacement:\n\n" .. tostring(span.suggested_replacement)
       end
 
       table.insert(diagnostics, {
@@ -37,7 +37,7 @@ return {
   parser = function(output, bufnr)
     local diagnostics = {}
     local items = #output > 0 and vim.split(output, "\n") or {}
-    local file_name = vim.fn.expand("%")
+    local file_name = vim.api.nvim_buf_get_name(bufnr)
 
     for _, i in ipairs(items) do
       local item = i ~= "" and vim.json.decode(i) or {}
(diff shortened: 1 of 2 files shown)


wispl:
Fixes some issues when using with pedantic mode. There can be multiple spans and each one is related to its parent message. Apparently clippy also outputs a suggested replacement message, currently I have it displayed by surrounding it with parenthesis, which might not be the best of ideas, especially when an empty string means deletion. Suggestions?

lua/lint/linters/clippy.lua
```diff
@@ -0,0 +1,51 @@
...
+  help = vim.diagnostic.severity.HINT,
+}
+
+local function parse(diagnostics, file_name, item)
+  for _, span in ipairs(item.spans) do
+    if span.file_name == file_name then
+      local message = item.message
+      if span.suggested_replacement ~= vim.NIL then
+        message = message .. " (" .. tostring(span.suggested_replacement) .. ")"
+      end
```
mfussenegger, line 13:
```suggestion
      if span.suggested_replacement ~= vim.NIL then
        message = message .. "\nSuggested replacement:\n\n" .. tostring(span.suggested_replacement)
      end
```

Not sure if that's better. I can also merge it as is.

lua/lint/linters/clippy.lua
```diff
@@ -0,0 +1,51 @@
...
+
+return {
+  cmd = "cargo",
+  args = { "clippy", "--message-format=json" },
+  stdin = false,
+  append_fname = false,
+  parser = function(output, bufnr)
+    local diagnostics = {}
+    local items = #output > 0 and vim.split(output, "\n") or {}
+    local file_name = vim.fn.expand("%")
```
mfussenegger, line 40:
```suggestion
    local file_name = vim.api.nvim_buf_get_name(bufnr)
```

Given that nvim-lint runs the linter async by the time the output gets parsed the active buffer could've changed, so this should use the `bufnr` to resolve the name.

wispl:
Applied the suggestions, I think the newlines might be better. Thanks for reviewing!

PR #630, opened 2024-08-09
Return table instead of a function for vsg

Anselmo95:
Refactor vsg.lua to return a table instead of a function as it prevents from changing the default configuration just partially.

In my case I just wanted to modify the `args` used by the linter but since `lint.linter.vsg` was a function I had to redefine it entirely. By returning a table i can just do the following in the config files:
```lua
      local ok, vsg = pcall(require, 'lint.linters.vsg')
      if ok then
        vsg.args = { <my args> }
      end
```

### DIFF
--- a/lua/lint/linters/vsg.lua
+++ b/lua/lint/linters/vsg.lua
@@ -39,8 +39,10 @@ local function find_config(dirname)
   end
 end
 
-return function()
-  local args = function()
+return {
+  cmd = "vsg",
+  stdin = true,
+  args = function()
     local args = { "-of", "syntastic", "--stdin" }
     local config_file = find_config(vim.fn.expand("%:p:h"))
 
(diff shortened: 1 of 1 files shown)


mfussenegger:
This will break the lazy resolving of `get_args`. It would get evaluated once on import and then be frozen, instead of getting evaluated on each use.

To modify the table you can wrap it in a function:

```lua
local l = require("lint")
local vsg = l.linters.vsg
l.linters.vsg = function()
  local linter = vsg()
  -- modify
  return linter
end
```

Anselmo95:
Thanks a lot for the answer, your solution works well but I seems a bit inconsistent with the example in the README.
```
local phpcs = require('lint').linters.phpcs
phpcs.args = {
  '-q',
  -- <- Add a new parameter here
  '--report=json',
  '-'
}
```
For vsg I cannot exactly do the same and change the arguments that's why i suggested returning a table, I took inspiration from the other files for this change.

PR #642, opened 2024-08-20
fix(sqlfluff): Disambiguate sqlfluff diagnostics

LJFRIESE:
Sqlfluff does not provide severity codes in its messages. The original defaulted everything to `ERROR`. This change downgrades regular messages to `WARN` and only flags the code for a parsing error as `ERROR`. This is done because Sqlfluff formatting will fail if a parsing error occurs, and identifying the error is much easier when a distinction is made between errors and warnings.


LJFRIESE:
Tests should pass now. They don't address the parse-failure that this change is disambiguating, but I'd rather not rewrite the whole test...

PR #643, opened 2024-08-21
docs: Document usage of ignore_errors option

AbelAnaya:
I added documentation for ignore_errors option of try_lint function. This is useful for users that want to lint a project only if the linter dependency is installed with the package manager locally, like it is usually done in JS and TS in other code editors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,33 @@ end
 ```
 
 
+## Ignore errors when invoking linters
+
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks but the readme was already getting longer than I'd like so I generated API docs instead: https://github.com/mfussenegger/nvim-lint/pull/655

Issue #519, opened 2024-01-22
Use npm installed linter when available


tobiabocchi:
Hey! thanks for all this work, I have been using your repo for a few days and I really like it!

I am trying to work with a js project which uses a specific version of `standardjs`, I noticed that it nvim-lint was not picking that one but the  one installed [system wide](https://github.com/mfussenegger/nvim-lint/blob/2cf9ad095130755d7d87f1730bcf33c91ee822e4/lua/lint/linters/standardjs.lua#L21) instead..

For formatting I am using conform.nvim which I think handles this correctly through a util method:

[here](https://github.com/stevearc/conform.nvim/blob/3d59cbd01a4b74124c5a6fb23b8cc63e5c2db3d5/lua/conform/formatters/standardjs.lua#L9) is the code that sets up `standardjs` as formatter, to parse the command to use it uses [this](https://github.com/stevearc/conform.nvim/blob/3d59cbd01a4b74124c5a6fb23b8cc63e5c2db3d5/lua/conform/util.lua#L6) util method.

I thought this might come in handy for a few linters in your project and maybe worth including.. I am not very experienced in lua, I mainly just use it to set up neovim.. If you don't have the time or don't think this is a priority I can try adding it myself and open a PR

AbelAnaya:
Hello!

I achieved this behavior by reusing some code of the `eslint` linter configuration in this project.

You can select how a linter behaves by modifying its properties in your configuration file for the plugin. An example for standard:

```lua
local lint = require("lint")
local binary_name = "standard"

local cmd_string = function()
    local local_binary = vim.fn.fnamemodify("./node_modules/.bin/" .. binary_name, ":p")
    return vim.loop.fs_stat(local_binary) and local_binary or binary_name
end

-- Use always local standard
local standardjs = lint.linters.standardjs
standardjs.cmd = cmd_string()
```

Simply add this code to your configuration function and `nvim-lint` will start using `standardjs` from local `node_modules` if possible, if not it will fallback to system wide installation.

I will try to add a PR that integrates the utils method of `conform` and uses it for `standardjs`in that project.

mfussenegger:
A PR that adapts standardjs to use the same pattern as eslint would be welcome.

But I'm starting wo wonder if this use-case wouldn't be in general better solved by adjusting your `$PATH` to something like `$PATH=./node_modules/.bin/:$PATH` - then you've consistend behavior across terminal and nvim-lint

casr:
[...]

PR #644, opened 2024-08-21
feat: Use npm installed package when possible

AbelAnaya:
I added two utils functions to search for commands to use as linter executable. These functions were ported from `conform.nvim` and have been adapted to `nvim-lint` as suggested in #519 

With these functions it is easy to setup linters that may have a local installation but are allowed to fallback to a system wide binary if no local is found.

I had no time to setup test execution, hopefully CI can do its work :). I manually tested that standardjs keeps working, so there shouldn't be any problem.

### DIFF
--- a/lua/lint/linters/standardjs.lua
+++ b/lua/lint/linters/standardjs.lua
@@ -17,8 +17,10 @@ local severities = {
   ['warning'] = vim.diagnostic.severity.WARN
 }
 
+local binary_name = "standard"
+
 return {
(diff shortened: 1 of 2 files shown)


mfussenegger (commented):
I'd rather have `standardjs` follow the pattern already used by some of the other linters like `stylelint` or `eslint`.

`lint.util` shouldn't have node specific functions, and `find_executable` looks like it would do way more than needed

PR #647, opened 2024-08-29
Correct GolangCI-Lint Installation Command

activcoding:
The current documentation incorrectly refers to the Go linter as `golangcilint`. Users attempting to install the linter with this name encounter the error: "couldn't find golangcilint."

This issue arises due to a missing hyphen in the command. The correct name of the linter is `golangci-lint`. This update rectifies the command to ensure that users can install the linter without encountering errors.


mfussenegger:
The linter name to use with `try_lint` or via `linters_by_ft` is `golangcilint` without hypen. Although it is unfortunate that it doesn't match the command name, changing this would be wrong.

PR #650, opened 2024-09-01
PHPCS linter fix

indahazzard:
I had problem with parsing messages, because output contains additional text, so I updated parser to pick only JSON part.

### DIFF
--- a/lua/lint/linters/phpcs.lua
+++ b/lua/lint/linters/phpcs.lua
@@ -22,26 +22,41 @@ return {
       return {}
     end
 
-    if not vim.startswith(output,'{') then
-      vim.notify(output)
+    local json_start = output:find("{")
+    local json_end = output:find("}%s")
+
+    if not json_start or not json_end then
+      vim.notify("No complete JSON found in output")
       return {}
     end
 
-    local decoded = vim.json.decode(output)
-    local diagnostics = {}
-    local messages = decoded['files']['STDIN']['messages']
+    local json_part = output:sub(json_start, json_end)
 
-    for _, msg in ipairs(messages or {}) do
-      table.insert(diagnostics, {
-        lnum = msg.line - 1,
-        end_lnum = msg.line - 1,
-        col = msg.column - 1,
-        end_col = msg.column - 1,
-        message = msg.message,
-        code = msg.source,
-        source = bin,
-        severity = assert(severities[msg.type], 'missing mapping for severity ' .. msg.type),
-      })
+    local success, decoded = pcall(vim.json.decode, json_part)
+    if not success then
+      vim.notify("Failed to decode JSON")
+      return {}
+    end
+
+    local diagnostics = {}
+
+    for file_path, file_data in pairs(decoded['files']) do
+      local messages = file_data['messages']
+
+      for _, msg in ipairs(messages or {}) do
+        print(msg.type)
+        table.insert(diagnostics, {
+          lnum = msg.line - 1,
+          end_lnum = msg.line - 1,
+          col = msg.column - 1,
+          end_col = msg.column - 1,
+          message = msg.message,
+          code = msg.source,
+          source = bin,
+          severity = assert(severities[msg.type],
+            'missing mapping for severity ' .. msg.type),
+        })
+      end
     end
 
     return diagnostics


mfussenegger (commented):
Can you elaborate and what kind of errors you got?

I'd expect a linter to only output JSON if the output is set to `JSON`, not some other errors. This might need to be reported upstream instead to get the linter fixed.

Software like nvim-lint should be able to use a regular json parser. Not custom logic to find json in some undefined output.

indahazzard:
Sure.
Here is output 
```
Registering sniffs in the PHP_CodeSniffer standard... DONE (105 sniffs registered)
Creating file list... DONE (0 files in queue)
{"totals":{"errors":0,"warnings":0,"fixable":0},"files":{}}
Time: 399ms; Memory: 8MB
```
Here is command I use to run it:
```
phpcs --standard=phpcs.xml -v --report=json
```

So I use `--standard` to define path to my file with standards
`-v` to get output
`--report=json` to get output in json format

Here is config I've tried to use to make it work:
```lua
return {
	"mfussenegger/nvim-lint",
	event = {
		"BufReadPre",
		"BufNewFile",
	},
	config = function()
		local lint = require("lint")


		lint.linters.phpmd.args = {
			'-',
			'json',
			vim.loop.cwd() .. '/phpmd.xml'
		}

		lint.linters.phpcs.args = {
			"-",
			"--standard=" .. vim.loop.cwd() .. "/phpcs.xml",
			"--report=json",
            "-q",
		}

		lint.linters_by_ft = {
			php = { "phpmd", "phpcs" }
		}

		local lint_augroup = vim.api.nvim_create_augroup("lint", { clear = true })

		vim.api.nvim_create_autocmd({ "BufEnter", "BufWritePost", "InsertLeave" }, {
			group = lint_augroup,
			callback = function()
				lint.try_lint()
			end
		})
	end

}
```

indahazzard:
Here is config I have now to make phpcs work:
```lua
return {
	"mfussenegger/nvim-lint",
	event = {
		"BufReadPre",
		"BufNewFile",
	},
	config = function()
		local lint = require("lint")


		lint.linters.phpmd.args = {
			'-',
			'json',
			vim.loop.cwd() .. '/phpmd.xml'
		}

		lint.linters.phpcs.args = {
			"-",
			"--standard=" .. vim.loop.cwd() .. "/phpcs.xml",
			"--report=json",
			"-v"
		}
		local severities = {
			['ERROR'] = vim.diagnostic.severity.ERROR,
			['WARNING'] = vim.diagnostic.severity.WARN,
			['INFORMATION'] = vim.diagnostic.severity.INFO,
			['HINT'] = vim.diagnostic.severity.HINT,
		}
		local bin = 'phpcs'
		lint.linters.phpcs.parser = function(output, _)
			if vim.trim(output) == '' or output == nil then
				return {}
			end

			local json_start = output:find("{")
			local json_end = output:find("}%s")

			if not json_start or not json_end then
				vim.notify("No complete JSON found in output")
				return {}
			end

			local json_part = output:sub(json_start, json_end)

			local success, decoded = pcall(vim.json.decode, json_part)
			if not success then
				vim.notify("Failed to decode JSON")
				return {}
			end

			local diagnostics = {}

			for file_path, file_data in pairs(decoded['files']) do
				local messages = file_data['messages']

				for _, msg in ipairs(messages or {}) do
					table.insert(diagnostics, {
						lnum = msg.line - 1,
						end_lnum = msg.line - 1,
						col = msg.column - 1,
						end_col = msg.column - 1,
						message = msg.message,
						code = msg.source,
						source = bin,
						severity = assert(severities[msg.type],
							'missing mapping for severity ' .. msg.type),
					})
				end
			end

			return diagnostics
		end

		lint.linters_by_ft = {
			php = { "phpmd", "phpcs" }
		}

		local lint_augroup = vim.api.nvim_create_augroup("lint", { clear = true })

		vim.api.nvim_create_autocmd({ "BufEnter", "BufWritePost", "InsertLeave" }, {
			group = lint_augroup,
			callback = function()
				lint.try_lint()
			end
		})
	end

}
```

indahazzard:
I think I will try to dive into the phpcs code. Maybe I am missing something because the '-q' option doesn't work for me, and I am not getting the JSON output, just some information about how long the command took to run and how much memory was used.

mfussenegger:
I assume the `-v` in `lint.linters.phpcs.args = { "-", "--standard=" .. vim.loop.cwd() .. "/phpcs.xml", "--report=json", "-v" }` implies verbose?
So maybe that's why you get extra output.


You can also check if the linter really outputs the extra info on the same stdio fd.
Tools often use stderr to show extra diagnostics while having the parseable output on stdout.

You can test that running something like `mycmd 2> /dev/null`

indahazzard:
Yes, `-v` stays for verbose. I've added `-v` because without it I have only Time and Memory as output.

indahazzard:
I closed because somehow my phpcs ignores '--report=json', so this is the reason why I don't see errors.
Sorry for bothering and thank you for you time.

PR #653, opened 2024-09-10
Add awk linter

fredrikfoss:
This PR adds a awk linter using (g)awk. I'm unsure if it instead should be called gawk, seeing as there are quite a few different AWK interpreters, and it's the gnu one that provides this linting functionality. I don't know about all the others. I saw [conform.nvim](https://github.com/stevearc/conform.nvim) use awk so I did as well.

Let me know what you think.


mfussenegger:
Thanks. I'm fine with just `awk`

PR #658, opened 2024-09-19
fix: format error in README example

usami:
The example config in Usage section has a format error.

This
```lua
  markdown = {'value',}
```
should be
```lua
  markdown = {'value'},
```


mfussenegger:
Thanks

PR #661, opened 2024-09-20
Add luac, ksh, and bash linters

eestrada:
I put each linter in its own commit so that it is easy to bisect the history and/or revert them individually.


mfussenegger:
Thanks

Issue #660, opened 2024-09-20
Filtering by path (`actionlint`)


9999years:
[`actionlint`](https://github.com/rhysd/actionlint) is a linter for GitHub Action workflows. However, if I include it in `linters_by_ft`, `nvim-lint` attempts to lint _all_ `yaml` files with `actionlint`, leading to copious errors if I open any non-GitHub Actions `yaml` file.

Proposal: Let's add another key to `lint.Linter` for filtering whether or not the linter is run on a particular file. Maybe it could get passed the buffer's path, or even just the buffer number and individual linters could resolve it from there.

Keeping the interface simple (a function you pass a buffer number to) will help keep complexity out of the core and push it into the individual linters, hopefully reducing maintenance burden. (I'm pretty mindful of the maintenance burden of the linters after `null-ls` got archived largely for that reason...)

I'd be happy to write the code for this, but I don't want to put in the effort if you're not on board with the design.

mfussenegger:
You can instead do something like this:

```lua
vim.filetype.add({
  pattern = {
    ['.*/.github/workflows/.*%.yml'] = 'yaml.ghaction',
    ['.*/.github/workflows/.*%.yaml'] = 'yaml.ghaction',
  },
})
```

```lua
  local lint = require('lint')
  lint.linters_by_ft = {
    ghaction = {"actionlint"},
  }
```

PR #663, opened 2024-09-20
Document how to use glob filters to only run linters on a subset of a filetype

9999years:
This documents the (excellent!) workaround for #660 in the `README.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,28 @@ end
 ```
 
 
+## Only run linters on certain glob patterns
+
+You might want to only run linters like [`actionlint`][actionlint] on a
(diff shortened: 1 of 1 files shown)


xudyang1:
For lazy loading nvim-lint, the following config may be helpful:

```lua
if filetype == "yaml" and string.find(path, ".github/workflows/") then
  set_linters_cwd(opts.root_patterns_by_ft.github_action)
  lint.try_lint(opts.linters_by_ft.github_action)
  return
end
```

mfussenegger:
Thanks, I integrated this a bit differently: https://github.com/mfussenegger/nvim-lint/pull/736

PR #665, opened 2024-09-23
feat: add api-linter (powered by buf CLI)

fredrikaverpil:
This adds the excellent "api-linter" from Google, which lints for so-called [AIPs](https://google.aip.dev/general) in `.proto` files: https://github.com/googleapis/api-linter

image

### Notes

* You must have `buf` on `$PATH`.
* There is a built in function (for the `--descriptor-set-in` argument) which attempts to search for a `buf.yaml` (or `buf.yml`) file. This search upwards from the opened buffer (the .proto file) until hitting `$HOME`, where it stops and fails. But this function may not be sufficient for everyone's use case. It's important that for such users, this function can be overridden and right now it seems like the only way is to fully replace all arguments with `opts.linters["api-linter"] = { args = {...} }`, which is totally fine I guess.
* The repo (and proto) from the screenshot: https://github.com/fredrikaverpil/go-microservice/blob/main/proto/gomicroservice/v1/user_service.proto
* This is how I use it in my Neovim/nvim-lint setup (but I would like to avoid all the manual setup and instead rely on implementation done here the nvim-lint repo): https://github.com/fredrikaverpil/dotfiles/blob/main/nvim-fredrik/lua/fredrik/plugins/lang/protobuf.lua
* I've chosen to include a few `--disable-rule` directives as otherwise you'll get these java warnings (and you may not even be using java). See below:

image


mfussenegger (commented):
Is this common to generate the descriptor-set-in file like this?

Looking at the other integrations mentioned in https://github.com/googleapis/api-linter/issues/764 they don't appear to be doing anything like this as far as I can tell.

fredrikaverpil:
@mfussenegger you have to somehow tell api-linter where it can find all imports that might be used by the proto file that you are linting, or api-linter will simply fail linting. I believe you can also do it via `-I` arguments, which you then need to point to other proto files (which correspond to your imports). One may traverse the current project for protos to solve that, but where would I traverse or look for imports from googleapis (downloaded buf dependencies)?
On my machine they are downloaded by the buf LSP into `~/.cache/buf` but I'm not sure that would be the case for everyone?
It would also take considerable amount of code to figure out whether to traverse that cache to find the appropriate protos, as it's not as simple as doing a `find` operation.

But by building this descriptor file, all such dependencies are sucked in (by buf itself) and redefined in the descriptor file, making it a lot easier to point api-linter to this derivative file. As a bonus here, buf will adhere to the buf lockfile, using the versions of the protos you intend.

So in my opinion I'm doing the right thing here. And obviously you can override the arguments and provide different logic if you don't think this suits your needs.

This is `api-linter --help` that describes `--descriptor-set-in` vs `-I`:

```bash
❯ api-linter --help
Usage of api-linter:
      --config string                   The linter config file.
      --debug                           Run in debug mode. Panics will print stack.
      --descriptor-set-in stringArray   The file containing a FileDescriptorSet for searching proto imports.
                                        May be specified multiple times.
      --disable-rule stringArray        Disable a rule with the given name.
                                        May be specified multiple times.
      --enable-rule stringArray         Enable a rule with the given name.
                                        May be specified multiple times.
      --ignore-comment-disables         If set to true, disable comments will be ignored.
                                        This is helpful when strict enforcement of AIPs are necessary and
                                        proto definitions should not be able to disable checks.
      --list-rules                      Print the rules and exit.  Honors the output-format flag.
      --output-format string            The format of the linting results.
                                        Supported formats include "yaml", "json","github" and "summary" table.
                                        YAML is the default.
  -o, --output-path string              The output file path.
                                        If not given, the linting results will be printed out to STDOUT.
  -I, --proto-path stringArray          The folder for searching proto imports.
                                        May be specified multiple times; directories will be searched in order.
                                        The current working directory is always used.
      --set-exit-status                 Return exit status 1 when lint errors are found.
      --version                         Print version and exit.
```

fredrikaverpil:
Here you can see from one of their tests how they provide knowledge about imports (other protos): https://github.com/googleapis/api-linter/blob/6ef36a69c0a9ff876e644922b413ea0999891c05/cmd/api-linter/cli_test.go#L19-L24

fredrikaverpil:
@mfussenegger I guess you can argue that this is solution is geared towards `buf` users. If you're not using `buf`, I'm not really sure how to solve this. But I think this is still a great starting point for people who _do_ use `buf`.

I use this almost everyday and it's super helpful when working with "AIP-driven" APIs: https://github.com/fredrikaverpil/dotfiles/blob/main/nvim-fredrik/lua/lang/protobuf.lua

fredrikaverpil:
@mfussenegger would you like me to implement two strategies, like;

1. `buf` CLI is available and `buf.yaml` is found; uses current implementation.
2. `buf` CLI may be available but `buf.yaml` is not found; walk `$cwd` for `*.proto` files and append each one using the `-I` argument.

Then I guess we can support what the other IDEs are doing too.

Strategy 2 `-I` will be inferior and will not support downloaded dependencies which reside outside your git repo project. But at least you will get something that looks similar to what other IDEs are doing.

mfussenegger:
No strong opinion on this. I don't use api-linter.

But it is highly unusual that a linter requires to run another tool beforehand. I kinda think that whatever `buf` is doing should be something that `api-linter` can do out of the box too.

The `descriptor_set_in` function also uses several functions only available in nvim 0.10+ and is synchronous.
Not a blocker either, but also not really happy about this.


If the buf setup is common, would it make sense to create a `buf-api-linter` project, that basically does the pre-processing and then delegates to `api-linter`?

mfussenegger:
> Sure, I guess that's one option, but in that case how would the "regular" nvim-lint api-linter would work by default?
> Would it just walk the $CWD recursively for .proto files and append them (using the -I argument) to the command?

Yes I I'd prefer that, plus the `buf-api-linter` variant. (Or only the `buf-api-linter` variant if the above behavior is not all that useful)

fredrikaverpil:
I think for most users, a "bare-bones" api-linter which wouldn't take any third-party plugins into account wouldn't be usable, really. So I opted for just adding a `buf_api_linter` here.

I also swapped out some newer functions for older and custom approaches. I'm not sure what you think about that. Anyway, I tested these changes out here on a couple of projects and it works fine for me.

mfussenegger:
I think there was a bit of a misunderstanding. What I meant was not only changing this to buf_api_linter but also create a dedicated buf-api-linter project with an executable that combines `buf` and `api-linter`.

Then this integration could define the `buf-api-linter` cmd and remove the logic around calling `buf`.

This would have the advantage that integration in other editors would make use of it too. 
(It's to me generally a bit of a miracle that we haven't managed to standardize output formats but need custom logic for each and every linter)

fredrikaverpil:
Aha okay. Yes, I think I misunderstood you there.

So `buf-api-linter` should call `try_lint("api-linter")` but with custom logic around `buf` and with custom opts?

bheadwhite:
bump

fredrikaverpil:
@mfussenegger please have a look at what I did now. I broke apart the logic into two linters.

I know the interface doesn't specify supplying an example autocmd, but I felt I had to do this to show the usage intent due to the lack of [cwd-as-a-function](https://github.com/mfussenegger/nvim-lint/pull/674).

mfussenegger:
See https://github.com/mfussenegger/nvim-lint/pull/665#issuecomment-2418986441

PR #668, opened 2024-09-30
fix(verilator): added file group

DinDjarins:
Issue: Earlier diagnostics of all files(in the included directories) for a line number was shown in the currently open buffer. Fix: Now the behaviour is normal since nvim-lint can now detect the file name and will show diagnostics relevant to the current buffer only.


DinDjarins:
I am not sure why the tests are failing.

mfussenegger:
Thanks. I added a fixup to fix the tests. The issue was that the filename in the test case no longer matched and the cwd argument is required if a file capture is present.

Issue #646, opened 2024-08-29
Set the cwd for each linter


lopi-py:
Let's say I have a project with the following structure:
```
project/
├── src/
│   ├── init.lua
│   └── foo.lua
├── selene.toml
└── subproject/
    ├── src/
    │   ├── init.lua
    │   └── bar.lua
    └── selene.toml
```
Currently, `nvim-lint` will set the cwd to the Neovim cwd, `project` in this case, but when trying to edit `subproject/src/init.lua`, the cwd should be `project/subproject` so selene will pick the correct configuration file

My proposal is to add a `cwd` field to the linter config
```lua
return {
  cmd = "selene",
  ...
  cwd = function(config, bufnr)
    return require("lint.util").root(bufnr, "selene.toml")
  end,
}
```
Would you be open to a PR to address this?

xudyang1:
```lua
local opts = {
  events = {
    -- loading buffer after nvim-lint: event="BufReadPost"
    "BufReadPost",
    -- some linters require saving file
    "BufWritePost",
    "InsertLeave",
    -- more aggresive (e.g., undo)
    "TextChanged",
  },
  linters_by_ft = {
    lua = {
      "selene",
      -- "luacheck",
    },
    github_action = {
      -- Static checker for GitHub Actions workflow files
      "actionlint",
      -- general yaml linter
      "yamllint",
    },
  },
  root_patterns_by_ft = {
    lua = {
      selene = { "selene.toml", ".git" },
      -- luacheck = { ".luacheckrc", ".git" },
    },
    github_action = {
      yamllint = { ".yamllint.yaml", ".yamllint.yml", ".yamllint" },
    },
  }  
}

local lint = require("lint")
local api = vim.api
local fs = vim.fs
lint.linters_by_ft = opts.linters_by_ft

---Set cwd for linters specified by opts.root_patterns_by_ft[filetype]
---@param linter_root_patterns table<string, string[]>? a mapping of linter to patterns
local function set_linters_cwd(linter_root_patterns)
  for linter, root_patterns in pairs(linter_root_patterns or {}) do
    local root_path = root_patterns and fs.root(0, root_patterns)
    if root_path then
      lint.linters[linter].cwd = root_path
      -- you can set cwd to buffer's parent dir if no root is found
      -- else
      --   lint.linters[linter].cwd = fs.dirname(api.nvim_buf_get_name(0))
    end
  end
end

[...]

PR #674, opened 2024-10-03
Allow linter's cwd to be specifed as function

fredrikaverpil:
### Why this change?

When I set per-linter `cwd` I'm noticing that the value gets evaluated too early, when Neovim is launched while the lazy package manager is assembling all plugins' `opts`. This becomes problematic as I have some logic that needs to find a certain configuration file in order to give me back the desired path for the `cwd`. And it's also only applicable if I open a certain file type in a neovim buffer.

So, I'd like to instead provide `cwd` as a function, so that it will be called when actually performing the linting and only for relevant file types.

To better explain what I mean, I define my nvim-lint linters in per-language lua config files, and then I have a main nvim-lint lua file which merges all the opts before actually passing the opts into nvim-lint:

* python: https://github.com/fredrikaverpil/dotfiles/blob/21b3ff2f2afad620ffc8e6060d1f46f8a9dc1638/nvim-fredrik/lua/lang/python.lua#L177
* go: https://github.com/fredrikaverpil/dotfiles/blob/65a48180f2e2543eb476cba0094f7ef1cc174c4e/nvim-fredrik/lua/lang/go.lua#L62
* protobuf: https://github.com/fredrikaverpil/dotfiles/blob/91213fc3050dab919645714479c771c0b9550f65/nvim-fredrik/lua/lang/protobuf.lua#L68
* lint.lua assembling all linters: https://github.com/fredrikaverpil/dotfiles/blob/21b3ff2f2afad620ffc8e6060d1f46f8a9dc1638/nvim-fredrik/lua/plugins/lint.lua#L16

You can see how I need to specify the `cwd` for protobuf and I don't want that logic to run for any other linter.

### What was changed?

- Add the capability to handle a linter's `opts.cwd` in case a function was passed in.

### Notes

- Should probably add a note about this in the docs txt and/or README?
- I'm a little confused by https://github.com/mfussenegger/nvim-lint/issues/646 which seems to indicate setting the `cwd` per linter is not even possible (or desired)... It feels like I've misunderstood something quite fundamental. But the docs mention it [here](https://github.com/mfussenegger/nvim-lint/blob/27f44d1cc3d733a38a736acb902f94879d99c76c/doc/lint.txt#L84) and `uv.spawn` does receive the `cwd` as part of the `linter_opts` passed into it ([here](https://github.com/mfussenegger/nvim-lint/blob/968a35d54b3a4c1ce66609cf80b14d4ae44fe77f/lua/lint.lua#L366)). It _seems_ to me I can set `cwd` per-linter just fine... and this PR change _seems_ to also work... 🤔

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -315,6 +315,9 @@ local function with_cwd(cwd, fn, ...)
   if curcwd == cwd then
     return fn(...)
   else
+    if type(cwd) == 'function' then
+      cwd = cwd()
+    end
     local mods = { noautocmd = true }
     vim.cmd.cd({cwd, mods = mods})
     local ok, result = pcall(fn, ...)


mfussenegger:
> I'm a little confused by https://github.com/mfussenegger/nvim-lint/issues/646 which seems to indicate setting the cwd per linter is not even possible (or desired).

yeah I kinda forgot that I added support for this, and also regret it a bit. Linter definitions should be mostly static & shareable between users and the cwd is more of a dynamic thing as it depends on the environment.


Looking at your configuration I wonder if this wouldn't be better solved by setting up different `try_lint` triggers.

E.g. I do have a generic:

```lua
  create_autocmd({'BufWritePost', 'BufEnter'}, {
    group = api.nvim_create_augroup('lint', { clear = true }),
    callback = function() lint.try_lint() end,
  })
```

For the linters defined in `linters_by_ft` but I also set up ad-hoc cmds like the following in filetype plugins:

```lua
local checkstyle_config = vim.uv.cwd() .. "/checkstyle.xml"
local has_checkstyle = vim.uv.fs_stat(checkstyle_config) and vim.fn.executable("checkstyle")
local is_main = api.nvim_buf_get_name(0):find("src/main/java") ~= nil
if has_checkstyle and is_main then
  api.nvim_create_autocmd({"BufEnter", "BufWritePost"}, {
    buffer = bufnr,
    group = api.nvim_create_augroup("checkstyle-" .. bufnr, { clear = true }),
    callback = function()
      if not vim.bo[bufnr].modified then
        require("lint.linters.checkstyle").config_file = checkstyle_config
        require("lint").try_lint("checkstyle")
      end
    end
  })
end
```

With that approach you could pass along the `{ cwd = ... }` options.

fredrikaverpil:
Thanks, that works. The `buf_lint_cwd()` call is only executed when I enter or save a `*.proto` file with this:

```lua
      vim.api.nvim_create_autocmd({ "BufEnter", "BufWritePost" }, {
        pattern = { "*.proto" },
        callback = function()
          require("lint").try_lint("buf_lint", {
            args = buf_lint_args,
            cwd = buf_lint_cwd(),
            append_fname = false,
          })
        end,
      })
```

However, since `cwd` is already supported/implemented in the `opts`, I don't quite understand why it wouldn't be okay to support `cwd` as a function (as proposed in this PR), without requiring this kind of workaround.

Either way, thanks for your hard work on nvim-lint and this suggestion! 😄

fredrikaverpil:
@mfussenegger I just saw you merged #735 and rebased this PR (and tweaked the change). Do you think something like this could work?

mfussenegger:
I still kinda regret that `cwd` was added to the linter table in the first place. So I don't really want to extend this further.
There have also been no upvotes on the PR, so I get the impression that not many want to have this either.

Closing it for now given that there are other options to accomplish your use-case

PR #676, opened 2024-10-04
feat: capture mypy error codes

tahv:
Capturing mypy [error codes](https://mypy.readthedocs.io/en/stable/error_code_list.html) to use in diagnostics.

image


mfussenegger:
Thanks

PR #688, opened 2024-10-24
fix: don't block UI when canceling lint process

stevearc:
I've been getting mysterious nondeterministic freezes in Neovim for months now, but couldn't get a solid repro. I finally found a repro and tracked it down to `LintProc:cancel()`. In #522 we added some logic to cancel a process by killing it with `sigint`, waiting, then if it's still running kill it again with `sigkill`. The problem is that the delay is done with `vim.wait()`, which will block the UI thread. If you have a poorly-behaved linter (and I do, unfortunately it's for work and not something I can reasonably fix) that doesn't respond at all to `sigint`, the UI thread will be blocked for 10 seconds each time it fails to cancel.

I've replaced the synchronous `vim.wait` with a call to `vim.defer_fn`. From comments on the previous PR I gather that there is some concern about orphaned lint processes when Neovim exits. There's not much we can do if it crashes, but for normal operation I added a `VimLeavePre` hook that will cancel all running linters and block until they are killed.

## Test Plan

First I found a repro without this patch
1. Open file
2. Call `lint.try_lint()` twice
3. Observe 10 second UI freeze

Then after applying this patch
1. Open file
2. Call `lint.try_lint()` twice
3. Observe no freeze
4. Run `ps aux | grep <formatter>` and verify that there are two processes running
5. Wait 10 seconds
6. Observe that the canceled process (the first one) was killed

And to check the exit behavior
1. Open file
2. Call `lint.try_lint()`
3. `:q`
4. Run `ps aux | grep <formatter>` and verify that the formatter is running
5. Observe 10 second UI freeze
6. Observe that Neovim quits and `ps aux | grep <formatter>` shows that the formatter process has been killed

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -5,6 +5,7 @@
 local uv = vim.loop
 local api = vim.api
 local notify = vim.notify_once or vim.notify
+local SIGKILL_WAIT_MS = 10000
 local M = {}
 
 --- Running processes by buffer -> by linter name
@@ -132,9 +133,18 @@ function M.get_running(bufnr)
   return linters
 end
 
-
-
-
+---@return integer
+local function count_running_procs()
+  local num_procs = 0
+  for _, linters in pairs(running_procs_by_buf) do
+    for _, proc in pairs(linters) do
+      if proc.handle and not proc.handle:is_closing() then
+        num_procs = num_procs + 1
+      end
+    end
+  end
+  return num_procs
+end
 
 ---Table with the available linters
 ---@type table<string, lint.Linter|fun():lint.Linter>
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):
The entries in the `running_procs_by_buf`  table get replaced after a `:cancel` call, so I think the `VimLeavePre` logic wouldn't work like this because the table won't contain the cancelled LintProcs anymore.

But the bigger problem is that if the cancel is deferred, you end up running multiple linters in parallel and you might get results updated/replaced differently.

Not sure how to deal with this. Maybe other options could be:

- Wrap your misbehaving linter in a small script that forwards sigint as sigkill
- `try_lint` could have a difference debounce mode, where instead of cancelling a proc, it lets it finish instead of starting a new process. Although this means the results would be stale 
- New lint process gets chained onto the previous lint process, so it only starts once the previous one got cancelled - but this is probably more complex than I'd like.

stevearc:
> The entries in the running_procs_by_buf table get replaced after a :cancel call, so I think the VimLeavePre logic wouldn't work like this because the table won't contain the cancelled LintProcs anymore.

That should be fine, right? Because we only need to cancel the currently running tasks on exit. Unless you're worried about the case of

1. Run lint process
2. Manually cancel lint process
3. Exit vim immediately
4. Vim exits before the sigkill timeout is reached, so process is orphaned

If that's what you're concerned about, I can fix it pretty easily by just having another way to keep track of all active LintProcs. LMK if this is what you were thinking of or if I'm misinterpreting.

> But the bigger problem is that if the cancel is deferred, you end up running multiple linters in parallel and you might get results updated/replaced differently.

I don't think this can happen because when a LintProc is cancelled it won't publish the (now stale) diagnostics. 

https://github.com/mfussenegger/nvim-lint/blob/ab83154f999b93bfbc99856e83337a1e5cf72809/lua/lint.lua#L242-L244

mfussenegger:
Merged the more minimal variant from https://github.com/mfussenegger/nvim-lint/pull/746 for now

stevearc:
Good enough for me, thanks!

PR #692, opened 2024-10-30
fix(protolint): check for vim.NIL

fredrikaverpil:
### Why this change?

After debugging an nvim-lint/protolint crash I got in a proto file, it turns out that the `json_output.lints` can be `vim.NIL` (not `nil`).

### What was changed?

Extend the nilcheck with checking for `vim.NIL`.

### DIFF
--- a/lua/lint/linters/protolint.lua
+++ b/lua/lint/linters/protolint.lua
@@ -12,7 +12,7 @@ return {
     end
     local json_output = vim.json.decode(output)
     local diagnostics = {}
-    if json_output.lints == nil then
+    if json_output.lints == nil or json_output.lints == vim.NIL then
       return diagnostics
     end
     for _, item in ipairs(json_output.lints) do


lua/lint/linters/protolint.lua
```diff
@@ -12,7 +12,7 @@ return {
     end
     local json_output = vim.json.decode(output)
     local diagnostics = {}
-    if json_output.lints == nil then
+    if json_output.lints == nil or json_output.lints == vim.NIL then
```
mfussenegger, line 15: Just fyi: There's also `vim.json.decode(output, { luanil = { object = true }})` - which can be useful if `null` appears often in the json and one doesn't need to distinguish between absent properties/null.

fredrikaverpil: Oh, nice - I didn't know. Good to know, thanks!

PR #694, opened 2024-11-06
Pass relative path to terraform validate instead of changing cwd

brablc:
`terraform validate` starts validation from the cwd (unless `-chdir` is used), when we want to validate just some file (actually directory as it cannot validate just one file), instead of changing cwd we have to pass it as an argument.

### DIFF
--- a/lua/lint/linters/terraform_validate.lua
+++ b/lua/lint/linters/terraform_validate.lua
@@ -7,7 +7,7 @@ local terraform_severity_to_diagnostic_severity = {
 return function()
   return {
     cmd = 'terraform',
-    args = { "validate", "-json", vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":.:h") },
+    args = { 'validate', '-json', vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ':.:h') },
     append_fname = false,
     stdin = false,
     stream = 'both',
--- a/lua/lint/linters/terraform_validate.lua
+++ b/lua/lint/linters/terraform_validate.lua
@@ -7,12 +7,11 @@ local terraform_severity_to_diagnostic_severity = {
 return function()
   return {
     cmd = 'terraform',
-    args = { 'validate', '-json' },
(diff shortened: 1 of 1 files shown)


gegoune:
This does not seem correct.

After this change I am getting false positives for uninitialised modules:

```
This module is not yet installed. Run \"terraform init\" to install all modules required by this configuration.
```

Module is installed and running `terraform validate` with `-chdir` doesn't emit those false positives, also running with `cwd`, as prior to this change validates correctly.

Help page for `validate` does not mention passing in path they way this PR introduced:

```
$ terraform validate -h
Usage: terraform [global options] validate [options]

  Validate the configuration files in a directory, referring only to the
  configuration and not accessing any remote services such as remote state,
  provider APIs, etc.

  Validate runs checks that verify whether a configuration is syntactically
  valid and internally consistent, regardless of any provided variables or
  existing state. It is thus primarily useful for general verification of
  reusable modules, including correctness of attribute names and value types.

  It is safe to run this command automatically, for example as a post-save
  check in a text editor or as a test step for a re-usable module in a CI
  system.

  Validation requires an initialized working directory with any referenced
  plugins and modules installed. To initialize a working directory for
  validation without accessing any configured remote backend, use:
      terraform init -backend=false

  To verify configuration in the context of a particular run (a particular
  target workspace, input variable values, etc), use the 'terraform plan'
  command instead, which includes an implied validation check.

Options:

  -json                 Produce output in a machine-readable JSON format,
                        suitable for use in text editor integrations and other
                        automated systems. Always disables color.

  -no-color             If specified, output won't contain any color.

  -no-tests             If specified, Terraform will not validate test files.

  -test-directory=path  Set the Terraform test directory, defaults to "tests".
```

Seems like passing in path is not doing what one would expect it to do, @brablc. However, as you mentioned, passing path with `-chdir` works fine.

I believe it should be changed to:

```
    args = { "-chdir=" .. vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":.:h"), "validate", "-json" },
```

What do you think?

PR #702, opened 2024-12-04
feat(ts-standard): Add ts-standard linter

AbelAnaya:
Add ts-standard to the list of supported linters


mfussenegger:
Could you please revert all the format changes in the readme?

PR #704, opened 2024-12-11
fix(linters): improve metatable handling for linter modifications

### LINKED ISSUE #703: Modifying a builtin linter without reduced form consider it as a new linter
```lua
{
  "nvim-lint",
  after = function()
    local lint = require("lint")
    lint.linters_by_ft = { nix = { "statix" } }
    lint.linters.fake_statix = {
      args = { "check", "-o", "errfmt", "--stdin" },
      cmd = "/nix/store/pkqp4is8aybllw37406hb3n0081jdhgq-statix-0.5.8/bin/statix",
      ignore_exitcode = true,
      parser = require("lint.parser").from_errorformat("<stdin>>%l:%c:%t:%n:%m", {
        source = "statix",
      }),
      stdin = true,
      stream = "stdout",
     }
     lint.linters.statix = { cmd = "/nix/store/pkqp4is8aybllw37406hb3n0081jdhgq-statix-0.5.8/bin/statix" }
  end,
  event = { "BufNewFile", "BufReadPost", "BufWritePost" },
 },
```

I've been setting nvim-lint with lz.n like this, but for some reason, I don't know if this is intentional, but when trying to modify statix without the reduced form (`lint.linters.statix.cmd`), the plugin considers that I am creating a new linter, just like I did with fake_statix, instead of modifying a builtin one.

HeitorAugustoLN:
Closes #703 

When modifying built-in linter properties using table assignment (e.g. lint.linters.name = { prop = value }), the plugin would create a new linter instead of modifying the existing one. This fix:

- Caches linters after first load for better performance
- Properly merges new properties with existing linter configurations
- Maintains existing linter functionality when modifying properties

BREAKING CHANGE: The behavior of linter modifications has changed. Direct assignments to lint.linters.name will now merge with existing linters instead of creating new ones.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -139,13 +139,29 @@ end
 ---Table with the available linters
 ---@type table<string, lint.Linter|fun():lint.Linter>
 M.linters = setmetatable({}, {
-  __index = function(_, key)
+  __index = function(tbl, key)
     local ok, linter = pcall(require, 'lint.linters.' .. key)
     if ok then
+      -- Cache the linter once loaded
+      rawset(tbl, key, linter)
       return linter
     end
     return nil
   end,
+  __newindex = function(tbl, key, value)
+    -- Try to load existing linter first
+    local ok, existing = pcall(require, 'lint.linters.' .. key)
+    if ok and type(value) == "table" then
+      -- If linter exists and we're setting a table, merge with existing
+      for k, v in pairs(value) do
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks but in lua `require` calls are already cached. All that's doing is use more memory.
And that assignments would merge configs is imho confusing - besides being a breaking change.

PR #726, opened 2025-01-18
feat: Add support to override terraform command to support OpenTofu

emolitor:
There is probably a better way to do this but pushing a quick change I'm using locally upstream for consideration. Feedback gratefully appreciatesd.

### DIFF
--- a/lua/lint/linters/terraform_validate.lua
+++ b/lua/lint/linters/terraform_validate.lua
@@ -4,9 +4,13 @@ local terraform_severity_to_diagnostic_severity = {
   notice = vim.diagnostic.severity.INFO,
 }
 
+-- Allow ability to override of terraform command to support OpenTofu
+-- vim.g.nvim_lint_terraform = "tofu" -- set in options.lua or init.lua
+local terraform = vim.g.nvim_lint_terraform or "terraform"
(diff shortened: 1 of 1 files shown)


mfussenegger:
This should either be a dedicated linter or done via the regular customization: https://github.com/mfussenegger/nvim-lint#customize-built-in-linters

E.g. a `opentofu.lua` could look like this if the only difference is the `cmd` argument:

```lua
return function ()
  local terraform = require("lint.linters.terraform_validate")()
  local opentofu = vim.deepcopy(terraform)
  opentofu.cmd = "opentofu"
  return opentofu
end
```

emolitor:
> This should either be a dedicated linter or done via the regular customization: https://github.com/mfussenegger/nvim-lint#customize-built-in-linters
> 
> E.g. a `opentofu.lua` could look like this if the only difference is the `cmd` argument:
> 
> ```lua
> return function ()
>   local terraform = require("lint.linters.terraform_validate")()
>   local opentofu = vim.deepcopy(terraform)
>   opentofu.cmd = "opentofu"
>   return opentofu
> end
> ```

Thank you for the quick 👁️ I will rework the change.

PR #735, opened 2025-02-06
Run all argument evaluations with cwd = opts/linter cwd

mfussenegger:
Alternative to https://github.com/mfussenegger/nvim-lint/pull/605


lua/lint.lua
```diff
@@ -305,6 +305,27 @@ local function eval_fn_or_id(x)
...
+    local mods = { noautocmd = true }
+    vim.cmd.cd({cwd, mods = mods})
+    local ok, result = pcall(fn, ...)
+    vim.cmd.cd({curcwd , mods = mods})
+    if ok then
+      return result
+    end
+    error(result)
+  end
+end
```
fredrikaverpil, line 327: Just a thought... would it have been better to call `vim.fn.chdir()` instead, which would change directory only for the current window? 🤔

PR #740, opened 2025-02-20
Add graylint adapter

noelevans:
Python has a linter called graylint. The tool uses other linters to assess the code (pylint, mypy, flake8) but then only flags lines that are new to git. Graylint returns line-by-line lint issues rather than json or other alternatives used by pylint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Other dedicated linters that are built-in are:
 | [gitlint][gitlint]                     | `gitlint`              |
 | [glslc][glslc]                         | `glslc`                |
 | [Golangci-lint][16]                    | `golangcilint`         |
+| [graylint][graylint]                   | `graylint`             |
 | [hadolint][28]                         | `hadolint`             |
 | [hledger][hledger]                     | `hledger`              |
 | [hlint][32]                            | `hlint`                |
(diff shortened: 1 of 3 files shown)


noelevans:
Thinking about this change more, I see that I'm hardcoding Graylint to the Pylint. Looking more at Ruff I see that it has many benefits over Pylint (which still seems to be the more established of the options Graylint can run). What is the best thing to do? Should I label this linter as `graylint-pylint` and make another for `graylint-ruff`? Graylint is also able to run mypy and flake8 internally too. What approach do you prefer @mfussenegger?

mfussenegger:
> Should I label this linter as graylint-pylint and make another for graylint-ruff? Graylint is also able to run mypy and flake8 internally too

Could be an option to remove the `args` and have users customize linter.
But overall in regards to https://github.com/mfussenegger/nvim-lint/issues/741 I'm not sure if this linter even needs to be included.

noelevans:
Yes, I think you are right. Rename to graylint-pylint and another for
graylint-ruff.

mfussenegger:
> But overall in regards to https://github.com/mfussenegger/nvim-lint/issues/741 I'm not sure if this linter even needs to be included.

Closing this here, I think I'd rather go the route of documenting ways the exclude results that aren't part of a current diff

PR #746, opened 2025-02-25
fix: do not block UI when cancelling lint

PeterCardenas:
Stole this contribution from [this PR](https://github.com/mfussenegger/nvim-lint/pull/688) since I do not have permissions to push to that branch and do not wish to continue maintaining my fork with this change. Minimized the contribution so that we only do the `vim.defer_fn` instead of `vim.wait` to remove the complexity of handling orphaned processes. As a side note, the handling of the orphaned processes on `VimLeavePre` caused my editor to freeze when exiting, so the change was not desirable for me anyways.

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -284,16 +284,15 @@ function LintProc:cancel()
   -- This is mostly useful for when `cmd` is a script with a shebang.
   handle:kill('sigint')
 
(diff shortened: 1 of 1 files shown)


PeterCardenas:
i didn't consider that, fwiw the linters that most affect me are mypy, pylint, and buf lint.

PR #748, opened 2025-02-27
fix(eslint): use 'npx eslint' when local binary is missing

### LINKED ISSUE #747: eslint not working in nvim-lint, but eslint_d works
Hello,

I am using nvim-lint with the Lazy.nvim plugin manager. Below is how I load the plugin:

```lua
{
    "mfussenegger/nvim-lint",
    config = function()
      local lint = require("lint")
      local eslint = { "eslint_d" }

      lint.linters_by_ft = {
        javascript = eslint,
        typescript = eslint,
        javascriptreact = eslint,
        typescriptreact = eslint
      }

      vim.api.nvim_create_autocmd({ "BufWritePost", "BufReadPost", "InsertLeave", "TextChanged" }, {
        callback = function()
          lint.try_lint()
        end
      })
    end
  }
```

When the code contains `local eslint = { "eslint_d" }`, it works. When I change it to `local eslint = { "eslint" }`, it does not work.

However, when I test "eslint" with none-ls.nvim (which I don't want to use), it does work:

```lua
{
    "nvimtools/none-ls.nvim",
    dependencies = {
      "nvimtools/none-ls-extras.nvim",
    },
    config = function()
      require("null-ls").setup({
        sources = {
          null_ls.builtins.formatting.stylua,
          null_ls.builtins.formatting.prettier,
          require("none-ls.diagnostics.eslint")
        }
      })
      vim.keymap.set("n", "<leader>gf", vim.lsp.buf.format, {})
    end
  }
```

Expected Behavior
- eslint should work in nvim-lint, just like eslint_d does.

Possible Causes
- Am I missing something in my setup?
- Could this be a bug in nvim-lint?

Any help or clarification would be greatly appreciated. Thanks!

edit: I made a PR for this problem. I hope that the PR is correct.

markuskreukniet:
Fixes #747
Resolves #747
Closes #747

### Description
This fixes an issue where `nvim-lint` fails to find ESLint if it's not installed globally.
- Previously, the command used `"eslint"`, which fails if it's not in `PATH`.
- Now, it falls back to `"npx eslint"`, which correctly finds either the local or global version.

### Why This Fix?
- `npx eslint` works even if `eslint` is not globally installed.
- This prevents silent failures and makes `nvim-lint` more reliable.

### DIFF
--- a/lua/lint/linters/eslint.lua
+++ b/lua/lint/linters/eslint.lua
@@ -7,7 +7,7 @@ local severities = {
 return {
   cmd = function()
     local local_binary = vim.fn.fnamemodify('./node_modules/.bin/' .. binary_name, ':p')
-    return vim.loop.fs_stat(local_binary) and local_binary or binary_name
+    return vim.loop.fs_stat(local_binary) and local_binary or "npx " .. binary_name
   end,
   args = {
     '--format',


markuskreukniet:
No problem. Thank you for your response!

PR #755, opened 2025-03-09
fix(tflint): match path when .tf file is not in current dir

poorpy:
# What
* use ":." instead of ":t" to obtain relative file name.
* add "--recursive" to default args

# Why
When using tools such as [terragrunt](https://terragrunt.gruntwork.io/) my repo consists of both Terraform modules and .hcl files with env-specific configurations. i.e

```
├── live
│   └── dev
│       ├── rds
│           └── terragrunt.hcl
├── modules
│   ├── rds
│   │   ├── main.tf
│   │   ├── outputs.tf
│   │   ├── README.md
│   │   └── variables.tf
```

* If I open my nvim at the root of the project `tflint` won't find any files without `--recursive` or `--chdir=modules/rds` flags.
* While using either of flags `tflint` returns paths that are relative to current dir.

### Sample tflint output:

```
❯ tflint --format=json --chdir="modules/rds" | jq
{
  "issues": [
    {
      "rule": {
        "name": "terraform_required_version",
        "severity": "warning",
        "link": "https://github.com/terraform-linters/tflint-ruleset-terraform/blob/v0.10.0/docs/rules/terraform_required_version.md"
      },
      "message": "terraform \"required_version\" attribute is required",
      "range": {
        "filename": "modules/rds/main.tf",
        "start": {
          "line": 1,
          "column": 1
        },
        "end": {
          "line": 1,
          "column": 1
        }
      },
      "callers": []
    }
  ],
  "errors": []
}
```

### Open questions
Instead of using `--recursive` I considered sth along the lines of 
```lua
"--chdir=".. vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":.:h")
```
but I'm not really sure if this will work 100% of the times.

### DIFF
--- a/lua/lint/linters/tflint.lua
+++ b/lua/lint/linters/tflint.lua
@@ -6,7 +6,7 @@ local terraform_severity_to_diagnostic_severity = {
 
 return {
   cmd = "tflint",
-  args = {"--format=json"},
+  args = {"--format=json", "--recursive"},
   append_fname = false,
   stdin = false,
   ignore_exitcode = true,
@@ -14,7 +14,7 @@ return {
     local decoded = vim.json.decode(output) or {}
     local issues = decoded["issues"] or {}
     local diagnostics = {}
-    local buf_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":t")
+    local buf_path = vim.fn.fnamemodify(vim.api.nvim_buf_get_name(bufnr), ":.")
 
     for _, issue in ipairs(issues) do
       if issue.range.filename == buf_path then
--- a/spec/tflint_spec.lua
+++ b/spec/tflint_spec.lua
@@ -3,7 +3,7 @@ describe('linter.tflint', function()
     local parser = require('lint.linters.tflint').parser
     local bufnr = vim.uri_to_bufnr('file:///main.tf')
     local result = parser(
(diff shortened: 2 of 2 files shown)


mfussenegger:
@steve-mt, @FelipeLema does this look okay to you?

FelipeLema:
I couldn't find any docs on the `:.` expansion, even though it seems to work as intended

can anyone point to the docs on that one?

FelipeLema:
> I couldn't find any docs on the `:.` expansion, even though it seems to work as intended
> 
> can anyone point to the docs on that one?

https://neovim.io/doc/user/cmdline.html#filename-modifiers

checking out

FelipeLema:
shouldn't `cwd` be handled too? @poorpy

poorpy:
Hi @FelipeLema,
What exactly do you mean by handling `cwd`? Do you mean something along those lines?:
* user opens `nvim` at the root of the project (I'm assuming layout from my example)
* user issues any of `:cd|:tcd|:lcd` (or sth similar)
    1. user changes `cwd` to `modules` - diagnostics still work as expected
    2. user changes `cwd` to `live` - no diagnostics with `--recursive`, linter returns error message if using `--chdir=...` approach.
    
TBH I don't know how to solve case no. 2 and from what I've seen, a similar linter ([terraform-validate](https://github.com/mfussenegger/nvim-lint/blob/master/lua/lint/linters/terraform_validate.lua) just uses `--chdir` approach.
I can change my code to use `--chdir` to be consistent WDYT?

FelipeLema:
I wanted to avoid doing checks on files outside of current scope: if I have several groups-of-configs under the repo/project root I _probably_ want to just check the ones of the group I'm currently modifying

at the same time, now that we're talking about it I just realized, _the group_ (or directory or "group root") I'm currently modifying may not be obvious (in my case, we used a file markers following internal convention)

So let's disregard my question

FelipeLema (approved):

PR #759, opened 2025-03-24
docs: Add another linter customization example

wyatt-avilla:
It'd be helpful to demonstrate how to customize linters without overwriting their arguments.

I extended the linter customization snippet in the readme with an example of prepending to mypy's existing arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -384,6 +384,11 @@ phpcs.args = {
   '--report=json',
   '-'
 }
+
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks but I'd like to avoid that the readme turns into a Lua/vim-stdlib tutorial. It's written under the assumption that people have some familiarity with those.

PR #761, opened 2025-03-25
fix: update golangci-lint configuration for v2

### LINKED ISSUE #760: Golangci-lint V2 breaking changesa
With Golangci-lint V2, the command line options have changed.  An exact port of the existing options would be

```
args = { 
    'run',
    '--output.json.path=stdout',
    '--issues-exit-code=0',
    '--show-stats=false',
    '--output.text.print-issued-lines=false',
    '--output.text.print-linter-name=false',
    function()
      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":h")
    end
},
```

But `'--output.text.print-issued-lines=false'` and `'--output.text.print-linter-name=false'` are unneeded with the way the output works now.

JonnyLoughlin:
Fixes #760

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -5,35 +5,45 @@ local severities = {
   convention = vim.diagnostic.severity.HINT,
 }
 
+-- Gets the correct agruements to run based on the version of golangci-lint
+local getArgs = function()
+  local ok, output = pcall(vim.fn.system, { 'golangci-lint', 'version' })
+  if not ok then
+    return
+  end
+
+  -- The golangci-lint install script and prebuilt binaries strip the v from the version
+  --   tag so both strings must be checked
+  if string.find(output, 'version v1') or string.find(output, 'version 1') then
+    return {
+      'run',
+      '--out-format',
+      'json',
+      '--issues-exit-code=0',
+      '--show-stats=false',
+      '--print-issued-lines=false',
+      '--print-linter-name=false',
+      function()
+        return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":h")
+      end
+    }
+  end
+
+  return {
+    'run',
+    '--output.json.path=stdout',
+    '--issues-exit-code=0',
+    '--show-stats=false',
+    function()
+      return vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ":h")
+    end,
+  }
+end
+
 return {
   cmd = 'golangci-lint',
   append_fname = false,
-  args = (function()
-    if string.find(vim.fn.system { 'golangci-lint', 'version' }, 'version v2') then
-      return {
-        'run',
-        '--output.json.path=stdout',
-        '--issues-exit-code=0',
-        '--show-stats=false',
-        function()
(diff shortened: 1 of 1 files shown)


oleynikandrey:
It only works for me if I set `--output.text.path=stderr`.

aNickPlx:
The fix has worked for me as is using golangci-lint version `2.0.1`

sieuwerts:
> The fix has worked for me as is using golangci-lint version `2.0.1`

This was also the case for me, thanks.

sieuwerts:
> It only works for me if I set `--output.text.path=stderr`.

This was also the case for me, thanks.

pebbe:
Don't do this. See https://github.com/mfussenegger/nvim-lint/issues/760#issuecomment-2758985229

JonnyLoughlin:
> Don't do this. See [#760 (comment)](https://github.com/mfussenegger/nvim-lint/issues/760#issuecomment-2758985229)

Updated with this suggestion, thank you

JonnyLoughlin:
> It only works for me if I set `--output.text.path=stderr`.

Is this in addition to the json path or instead? The linter configuration is set to use stdout so it is strange that would be required. 

What platform are you on?

oleynikandrey:
> Is this in addition to the json path or instead? The linter configuration is set to use stdout so it is strange that would be required.

I found the reason: the output was redefined in the `.golangci.yaml` file. So, if someone wants to change it, for example, for CI, the plugin will be broken.

JonnyLoughlin:
> > Is this in addition to the json path or instead? The linter configuration is set to use stdout so it is strange that would be required.
> 
> I found the reason: the output was redefined in the `.golangci.yaml` file. So, if someone wants to change it, for example, for CI, the plugin will be broken.

The command line options should take precedence over the config file, so if the issue is defining 
```
output:
  formats:
    text:
      path: stdout
```
in the config file for CI, we should be able to just add the `--output.text-path-stderr` in the v2 args to make sure the linter doesn't write the text to stdout when run by nvim-lint.

If that sounds correct, let me know and I can update the fork.

oleynikandrey:
> If that sounds correct, let me know and I can update the fork.

Yes, this will suffice.

JonnyLoughlin:
Does anyone have input on if we should set:
```
-- Overwrite value possibly  set in .golangci.yml
'--output.text.path=""',
'--output.tab.path=""',
'--output.html.path=""',
'--output.checkstyle.path=""',
'--output.code-climate.path=""',
'--output.junit-xml.path=""',
'--output.teamcity.path=""',
--output.sarif.path=""',
```
or just 
```
'--output.text.path=""',
```

`--out-format json` previously would clear out any formats set in the config before setting the value to `json`, the which behaves differently with the new config structure.  I think we should clear out any other output formats to ensure the behavior shouldn't change if anyone upgrades from v1 to v2 and happens to have any of these set.

oleynikandrey:
This merge request is not complete because it does not clean the `output.text.path` parameter.

JonnyLoughlin:
> This merge request is not complete because it does not clean the `output.text.path` parameter.

Addressed in #770

PR #762, opened 2025-03-25
Enable using the wrapper as a filter

TimenZan:
Only a documentation/type change is needed. Setting a value in a table to nil already removes it, so no explicit filter is needed. This enables the usecase of removing a diagnostic entirely, rather than only modifying it.

For example:
```lua
lint.linters.languagetool =
    require("lint.util").wrap(lint.linters.languagetool, function (diagnostic)
        diagnostic.severity = vim.diagnostic.severity.HINT
        if diagnostic.message == "Possible typo: you repeated a whitespace" then
            return nil
        end
        return diagnostic
    end)
```

### DIFF
--- a/lua/lint/util.lua
+++ b/lua/lint/util.lua
@@ -36,7 +36,7 @@ end
 --- ```
 ---
 ---@param linter lint.Linter|fun():lint.Linter
----@param map fun(d: vim.Diagnostic): vim.Diagnostic
+---@param map fun(d: vim.Diagnostic): vim.Diagnostic|nil
 ---@return lint.Linter|fun():lint.Linter
 function M.wrap(linter, map)
   local function _wrap(l, m)


TimenZan:
Turns out that this can cause issues when publishing diagnostics, will debug and update later.

mfussenegger:
Having holes in lists can be problematic as it breaks stuff like `#`
To make this safe it should omit the nil entries in the result list:

```diff
diff --git a/lua/lint/util.lua b/lua/lint/util.lua
index 80c97c0..0f6bcdf 100644
--- a/lua/lint/util.lua
+++ b/lua/lint/util.lua
@@ -23,10 +23,13 @@ end
 
 --- Return a linter that wraps another Linter with a mapping function applied
 --- on its parser function. This allows to modify diagnostics generated by a
 --- linter.
 ---
+--- If the mapping function returns `nil` for a diagnostic, the original
+--- diagnostic is excluded.
+---
 --- Example, to remap the severity of a linter:
 ---
 --- ```lua
 --- local lint = require("lint")
 --- lint.linters.cspell = require("lint.util").wrap(lint.linters.cspell, function(diagnostic)
@@ -34,18 +37,25 @@ end
 ---   return diagnostic
 --- end)
 --- ```
 ---
 ---@param linter lint.Linter|fun():lint.Linter
----@param map fun(d: vim.Diagnostic): vim.Diagnostic
+---@param map fun(d: vim.Diagnostic): vim.Diagnostic?
 ---@return lint.Linter|fun():lint.Linter
 function M.wrap(linter, map)
   local function _wrap(l, m)
     local result = vim.deepcopy(l)
     result.parser = function(...)
       local diagnostics = l.parser(...)
-      return vim.tbl_map(m, diagnostics)
+      local mapped_diagnostics = {}
+      for _, d in ipairs(diagnostics) do
+        local mapped = m(d)
+        if mapped then
+          table.insert(mapped_diagnostics, mapped)
+        end
+      end
+      return mapped_diagnostics
     end
     return result
   end
   if type(linter) == "function" then
     return function()
```

mfussenegger:
Thanks for the initiative. Merged via https://github.com/mfussenegger/nvim-lint/pull/768

PR #767, opened 2025-04-02
feat: add support for fortitude

gomfol12:
Added a linter for fortitude.

I am not sure if the test is necessary as it is incredibly basic, but now it there.

Let me know if I can improve it.


mfussenegger:
Thanks

Issue #744, opened 2025-02-24
golangci-lint exit code 5 and 7


thesimonho:
I'm having a couple of different issues with golangci-lint. sometimes it will exit with code 5, and other times a code 7

Reproduce:
```
mkdir gotest
cd gotest
go mod init test
touch test.go
```
test.go:
```
package main

import "fmt"

func main() {
	error here
	fmt.Println("Hello, World!")
}
```
In the exit code 7 case, a warning will fire saying `Linter command golangci-lint exited with code: 7`, but the error diagnostics will still show in the buffer

I have a bigger Go project where I consistently get an exit code 5, but I cant seem to create a repro of it. In that case the error diagnostics in the buffer will slowly get out of sync with the buffer (ie ill fix an error, but the diag still remains)

I've confirmed in both cases that manually running `golangci-lint run` in the directory containing the file will work correctly

This was all working fine until ~2 months ago, so maybe related to these commits? https://github.com/mfussenegger/nvim-lint/commit/3d5190d318e802de3a503b74844aa87c2cd97ef0 https://github.com/mfussenegger/nvim-lint/commit/ba3a77b016afe2abeaf216b4c1960e9e5128747e

version: `golangci-lint has version 1.64.5 built with go1.24.0 from 0a603e49 on 20
25-02-13T21:19:55Z`

Pitasi:
I saw this behavior as well, and, for a split second, I think I saw a log saying that you can't have multiple golangci-lint instances running in parallel. Take it with a grain of salt though, might be unrelated.

Integralist:
I noticed the same thing but when running `golangci-lint run` manually I did notice an error right at the top of the output that indicated a config issue.

I've since opened this discussion:
https://github.com/golangci/golangci-lint/discussions/5683 which isn't directly related, but...might be?

brunofjesus:
I just installed this plugin and I noticed the same.
Then I found out that it only works fine if I open neovim in the same dir as the `go.mod` file and the `.golangci.yml`.

Integralist:
So I resolved my golangci-lint config issue but I was still seeing these errrors, and nothing is being reported in neovim.

I then noticed @brunofjesus comment about it only working if opening Neovim from the main go.mod directory, and that seemed to not show the error anymore.

But that said, golangci-lint still doesn't actually work for me :-/ 

For example, I have durationcheck setup in golangci-lint (see config here), and so running golangci-lint in the shell it'll tell me...

```
[...]

PR #771, opened 2025-04-04
fix: Error code 5 with golangci-lint v2

stefanbildl:
I had the same problems with exit code 5 using golangci-lint version 2 as mentioned in #744.

The comment by @brunofjesus https://github.com/mfussenegger/nvim-lint/issues/744#issuecomment-2776695600 helped me solve part of the issue.

I added code that goes up the directory tree to find the go mod file.
Also I added /... to the end of the path to handle some weird cases where golangci-lint still complained.

This seems to fix the issue with error code 5 for me. The problem with exit code 7 is still unresolved

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -5,6 +5,25 @@ local severities = {
   convention = vim.diagnostic.severity.HINT,
 }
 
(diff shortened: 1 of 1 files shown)


stefanbildl:
related to #761

Issue #772, opened 2025-04-06
golangci.yml file not in the cwd will cause the prompt to fail normally


tyanxie:
In order to allow all the go projects in my computer to use the .golangci.yml file directly instead of rewriting it every time, I put the .golangci.yml file in the _home_ directory, like `~/.golangci.yml`.

In version V1, golangci-lint worked well, but when I updated to version V2, it did not work properly.

After investigation, I found that the cause of the problem was that the file path output by golangci-lint v2 was relative to the .golangci.yml file, not relative to the cwd.

in golangci-lint v1:

<img width="1053" alt="Image" src="https://github.com/user-attachments/assets/acd00081-31f5-41e6-82a8-815b62084d94" />

in golangci-lint v2:

<img width="1010" alt="Image" src="https://github.com/user-attachments/assets/1c303b9e-9f8d-4742-94e3-4350ced0b36b" />

It's looks like when matching the file path, if golangci-lint version is v2, then it needs to be matched relative to the directory where the .golangci.yml file is located, not cwd.

I have tried to fix it using this method, and it works well in both v1 and v2 versions. I will make a pull request later. Hope you can give suggestions or merge it when you are free!

tyanxie:
I found a configuration that can modify the calculation method of relative paths. By default, it uses the path of the configuration file, but it can be changed to `cwd` (which is not recommended in the official documentation), and there seems to be no command line option to force the change to cwd.

In this case, should we go to suggest golangci-lint offical to add a `--relative-path-mode` option or should we calculate it according to the default cfg? Or do we need to parse the configuration file to obtain it dynamically (I think this is feasible, but it doesn't seem elegant)?

https://golangci-lint.run/usage/configuration/#run-configuration

```yaml
# Options for analysis running.
run:
  # The mode used to evaluate relative paths.
  # It's used by exclusions, Go plugins, and some linters.
  # The value can be:
  # - `gomod`: the paths will be relative to the directory of the `go.mod` file.
  # - `gitroot`: the paths will be relative to the git root (the parent directory of `.git`).
  # - `cfg`: the paths will be relative to the configuration file.
  # - `wd` (NOT recommended): the paths will be relative to the place where golangci-lint is run.
  # Default: cfg
  relative-path-mode: cfg
```

tyanxie:
Or maybe we can wait for the next release of golangci-lint to add the `--path-mode=abs` option in https://github.com/golangci/golangci-lint/pull/5651 ?

[...]

PR #773, opened 2025-04-06
feat(golangcilint): based on config file directory to match file in v2

tyanxie:
Fix https://github.com/mfussenegger/nvim-lint/issues/772


mfussenegger:
I think this is obsolete with https://github.com/mfussenegger/nvim-lint/pull/793 and https://github.com/mfussenegger/nvim-lint/pull/770 ?

PR #774, opened 2025-04-07
feat(linters): add redocly

prepor:
`redocly lint` is an [OpenAPI linter](https://redocly.com/docs/cli/commands/lint). The best format I found is `codeclimate`, although it doesn't return anything about the location except the line number. For OpenAPI specs, it's usually thou enough.


mfussenegger:

prepor:
done

PR #775, opened 2025-04-08
[FIX] oelint-adv: fix parse pattern to support oelint-adv `[branch:XXX]`

iliaik:
Starting from commit:
```
commit a620ab1cefdeeef35f2e17150534b31d9a362529
Author: Konrad Weihmann <kweihmann@outlook.com>
Date:   Tue Aug 27 12:18:24 2024 +0000
```
`oelint-adv` adds `[branch:XXX]` to the end of the output line regardless of the provided `--messageformat`. So the output needs to be parsed correspondingly as it contains extra ':' symbol now.


iliaik:
Please note that `_spec` file was changed blindly.

PR #786, opened 2025-05-05
fix(golangcilint): take config path into consideration when comparing file paths

pengux:
When there is a `golangci-lint.yaml` config file somewhere in path then the file paths that are returned from the results are all relative to the path of the config file.

This commit takes the config folder path into consideration when comparing the path of the file in vim with the path of the files returned from golangci-lint

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -49,6 +49,18 @@ local getArgs = function()
   }
 end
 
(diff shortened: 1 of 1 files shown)


mfussenegger:
I doubt this will work given that the command is executed on import time. If you switch between projects  the path will be wrong

pengux:
@mfussenegger is there any way to read the path of config when that happens? A vim event perhaps?

PR #790, opened 2025-05-13
feat(linters): add trivy_secret

BestChinchilla:
nvim-lint already have trivy linter, but it only scans for [misconfig](https://github.com/mfussenegger/nvim-lint/blob/fdb04e9285edefbe25a02a31a35e8fbb10fe054d/lua/lint/linters/trivy.lua#L12), this PR adding secret scanner functionality to make more safer development process

POC:
изображение


mfussenegger:
Would it be possible and make sense to extend the existing `trivy` linter to check both? 

cc @pbnj @bobsrac

BestChinchilla:
I can try adding more if conditions for each of the scanners, but I don't write in Lua, so it might not be exactly best practice code :D

PR #791, opened 2025-05-14
fix json decoding error in linters/pylint.lua

jerryfletcher21:
Fixes this error that prevented pylint from working: 
"Expected value but found invalid token at character 1"


mfussenegger:
What was the output? Does looks as if it is just swallowing legitimate errors.

jerryfletcher21:
@mfussenegger when I opened a .py file, pylint did not work, and the lsp reported kind of a stacktrace that pointed at the line that I changed with the error: "Expected value but found invalid token at character 1".
Doing this change pylint is working fine.

PR #793, opened 2025-05-19
fix(golangcilint): get absolute path from golangci-lint v2

cativovo:
The value from `lintedfile_norm` is giving a wrong path. We can directly get the absolute path of the linted file by using `--path-mode=abs` flag. This fix is based on none_ls' [implementation](https://github.com/nvimtools/none-ls.nvim/blob/90e4a27ccaa25979a6b732b9f06dfa43b54957b7/lua/null-ls/builtins/diagnostics/golangci_lint.lua#L40).

image

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -32,6 +32,11 @@ local getArgs = function()
   -- Omit --path-mode=abs, as it was added on v2.1.0
   -- Make sure it won't break v2.0.{0,1,2}
   if string.find(output, 'version v2.0.') or string.find(output, 'version 2.0.') then
+    -- If the linter is not working as expected, users should explicitly add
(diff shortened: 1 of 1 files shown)


mfussenegger:
@JonnyLoughlin does this look okay to you?

cativovo:
Sorry, I missed that case. I've updated the PR. Unfortunately, this fix won't resolve the issue on v2.0.{0,1,2}. The workaround is to add `run.relative-path-mode: wd` to their .golangci.yaml config, since the default value of `run.relative-path-mode` prior to v2.0.0 was `wd` https://golangci-lint.run/product/migration-guide/#runrelative-path-mode

JonnyLoughlin:
That should still be fine.  If anyone was using v2.0 before and it worked for them, this won't break their setup now.  Maybe just a final update to mention that workaround in a comment when v2.0 is found, otherwise I'd say this looks good to me.

PR #799, opened 2025-05-25
feat(phpstan): provide diagnostics with col start and end

nickkadutskyi:
PHPStan provides only line number of the error and no column so it is useful to provide diagnostics with col start and end of the actual content of the line to have better highlighting.


mfussenegger (commented):
Could `phpstan` be changed upstream to report end_col?

I think so far no other linter infers the `end_col` based on the keyword/whitespace.

Issue #482, opened 2023-12-05
Support monorepos through inferring sub-project root directory


GNRSN:
Hi,

I'm trying to setup nvim-lint to work in my monorepo projects.

My use case specifically is with webdev js/ts and eslint (eslint_d). Where the default behavior results in a *no config file found* error from eslint when no config file is present in the root of nvim cwd.

After looking at the source code of similar plugins (`null-ls`, `conform`, `lsp-config`) they employ strategies for finding (and caching) the corresponding project root for a file/buffer, see e.g. [usage from null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/0010ea927ab7c09ef0ce9bf28c2b573fc302f5a7/lua/null-ls/builtins/diagnostics/eslint.lua#L55) and [util in null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/0010ea927ab7c09ef0ce9bf28c2b573fc302f5a7/lua/null-ls/utils/init.lua#L254)

My workaround  at the moment reuses the util from lsp-config through `lspconfig.util`, searching for `package.json` to find root and overwriting the `litners.eslint_d.cwd` property before each`try_lint()` call. (Passing the cwd option to try lint isn't working for me)

Suggestion:
Dynamically find (and cache) cwd per linter per buffer.

mfussenegger:
Sorry but I currently have no plans to add something like that. That would get in the business of project detection/management, and there's lots of ways people handle that.

You could piggyback on the LSP machinery and query the root_dir of the lsp clients, and set the cwd based on that

caliguIa:
Anyone find a nice solution for this?

GNRSN:
> Anyone find a nice solution for this?

I just switched to [none-ls](https://github.com/nvimtools/none-ls.nvim) linting.

mfussenegger:
> You could piggyback on the LSP machinery and query the root_dir of the lsp clients, and set the cwd based on that

To give a concrete example:

```lua
local get_clients = vim.lsp.get_clients or vim.lsp.get_active_clients
local client = get_clients({ bufnr = 0 })[1] or {}
lint.try_lint(nil, { cwd = client.root_dir })
```

magoz:
Thanks @mfussenegger, that works!

walker-tx:
I had a bit more luck with this (for `ts_ls`):

```lua
local flex_lint = function()
	local client = get_lsp_client()
[...]

PR #800, opened 2025-05-26
feat:resharper

### LINKED ISSUE #798: [feature]: Support JetBrains Resharper linting for C#
I am currently working on this.

xentropic-dev:
Implements support for JetBrains resharper code inspection CLI tools.
Closes #798


lua/lint/linters/resharper.lua
```diff
@@ -0,0 +1,96 @@
...
+          code = result.ruleId,
+          filename = full_path,
+        })
+      end
+    end
+    return results
+  end
+
+  return {
+    cmd = "jb",
```
mfussenegger, line 78: Is this the official name? E.g. in the [AUR](https://aur.archlinux.org/packages/jetbrains-resharper-commandlinetools) package there is a `inspectcode` command. In the upstream .zip there are `inspectcode.exe` and `inspectcode.sh`. No `jb`.

It is also a bit strange to me that this doesn't support writing the output to either stdout or stderr like every other linter. Or that it can't discover the solution file by itself.

Maybe worth making some upstream PRs first to get this into a shape that makes it easier to integrate.

(And the PR can also be adapted to use the `for_sarif` parser function.

xentropic-dev: Btw, I'm really not sure the point of the AUR.  I use arch as well and managing the package via dotnet tool works great.  Not sure what circumstance people are in where they want to use Resharper without dotnet.  I'm sure there's a reason.

xentropic-dev: I got it.  I just think you stumbled upon an AUR that probably isn't widely used.  The normal way to install is with dotnet tool

mfussenegger: Okay based on the AUR conversation there are actually two different tools, so going with `jb` is fine. 

But I'd still like to see https://youtrack.jetbrains.com/issue/RSRP-500996 happen before getting this in. Of the ~150 linters included in nvim-lint, this would otherwise be the only one requiring a tempfile. 

I also wonder if this should be called simply `jb` instead of `resharper` to match the executable?

xentropic-dev: > But I'd still like to see https://youtrack.jetbrains.com/issue/RSRP-500996 happen before getting this in. Of the ~150 linters included in nvim-lint, this would otherwise be the only one requiring a tempfile.

Makes sense.  I'll keep track of the ticket and bump it if I don't see anything.

> I also wonder if this should be called simply `jb` instead of `resharper` to match the executable?

I can get behind this, especially since there are technically some subtle differences between the Resharper plugin and the Resharper command line tools (jb inspectcode).

Thanks!

xentropic-dev:
Also just noticed I accidentally reformatted the readme; i'll get that fixed.

lua/lint/linters/resharper.lua
```diff
@@ -0,0 +1,60 @@
...
+
+  if filepath == "" then
+    vim.notify("Current buffer has no file associated", vim.log.levels.WARN)
+    return {}
+  end
+
+  local solution_directory = vim.fn.fnamemodify(root_file, ":h")
+
+  local relative_filepath = filepath:sub(#solution_directory + 2)
+  local cache_directory = vim.fn.stdpath("cache") .. "/nvim-lint-resharper"
```
mfussenegger: Could this be changed to only search `.sln` or `.csproj` relative to the current dir?
Adding root dir heuristics as part of the linter is out of scope for nvim-dap. `cwd` can be parameterized as part of the `try_lint` call for users who have `cwd != project_dir`.

lua/lint/linters/resharper.lua
```diff
@@ -0,0 +1,60 @@
...
+
+  -- Ensure cache directory exists
+  vim.fn.mkdir(cache_directory, "p")
```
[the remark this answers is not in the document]
mfussenegger: If there wasn't any difference in your tests I'd be in favor of removing it.

lua/lint/linters/resharper.lua
mfussenegger: > So I guess it's fine as long the user always has csproj or sln in the cwd. For me, that rarely is the case because 

You could still do something as described in https://github.com/mfussenegger/nvim-lint/issues/482#issuecomment-1999185606

And at some point in the future there might be a solution based on https://github.com/neovim/neovim/issues/34622 

So for now I'd restrict resharper to only search downward from cwd

mfussenegger:
Closing for now as https://github.com/mfussenegger/nvim-lint/pull/800#discussion_r2625791287 wasn't addressed.

Please re-open/ping if that's addressed

PR #835, opened 2025-08-01
feat: adds fieldalignment linter

catgoose:
Adds linter parser for go fieldalignment.  This used to be part of the gopls LSP but it was removed in a recent version.

https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment

image

### DIFF
--- a/lua/lint/linters/fieldalignment.lua
+++ b/lua/lint/linters/fieldalignment.lua
@@ -11,7 +11,7 @@ return {
     "-json",
   },
   ignore_exitcode = true,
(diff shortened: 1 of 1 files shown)


mfussenegger (commented):

catgoose:
Added it to the readme

mfussenegger:
Judging by https://github.com/golangci/golangci-lint/discussions/2194 isn't this already included via the `golangcilint` linter?

Are there any distro packages that provide a `fieldalignment` command?

PR #848, opened 2025-09-23
Modify LintProc:publish method

obemu:
The method now uses the value of `vim.Diagnostic.bufnr`, if it is set, instead of always using `self.bufnr`. 

This feature is useful, when executing a linter once, returns lints for the entire working directory.
[See my nvim config for linting protobuf files](https://gitlab.com/obemu/init.lua/-/blob/master/after/plugin/20_setup_protobuf.lua).

### DIFF
--- a/lua/lint.lua
+++ b/lua/lint.lua
@@ -231,11 +231,32 @@ local linter_proc_mt = {
 }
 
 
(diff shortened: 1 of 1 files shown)


mfussenegger:
Thanks but supporting project level linting will need a bit more thought and a more holistic approach

PR #911, opened 2026-02-21
Fix detect secrets

LucasAVasco:
Fix 3 bugs related to detect-secrets linter:

- Executable and name. The correct detect-secrets executable is `detect-secrets` (instead of `linter-secrets`). See the official documentation at https://github.com/Yelp/detect-secrets

- Diagnostic column. Must be provided, otherwise Neovim throws an error. Because detect-secrets does not provide a column, I used a dummy column of 0

- Type annotation for diagnostic (must be `vim.Diagnostic` instead of `vim.Diagnostic.Set`)

### DIFF
--- a/lua/lint/linters/detect-secrets.lua
+++ b/lua/lint/linters/detect-secrets.lua
@@ -1,4 +1,5 @@
-local name = "linter-secrets"
+local name = "detect-secrets"
+
 return function()
   local args = { "scan" }
 


mfussenegger:
The last two commits shouldn't be necessary. The type annotations for `lint.parse` and `lint.Parser` could actually be changed to use `vim.Diagnostic.Set` instead of `vim.Diagnostc`.

Otherwise lgtm

LucasAVasco:
All right. I removed the last 2 commits.

PR #913, opened 2026-03-24
Allow for a function to be passed in ``linter.args`` instead of just a list.

thepigeongenerator:
See patch commits & their messages for more information.
The first patch contains a fix for Windows systems, regarding duplicating the ``linter.args`` value. The second contains the actual change. Finally, the last updates the README to provide some (minor) documentation for this feature.


mfussenegger:
Any reason returning the full linter as function is not sufficient? That also allows generate all args dynamically.

thepigeongenerator:
I saw so a moment ago, it is. (this is also the reason why I returned here)
I will leave the pull request open, since it may be something you'd like to add.
Otherwise, it is also good to leave it here.
Thanks!

mfussenegger:
Thanks for the reply. I'll keep referring to returning the linter as function for now to keep it simpler

PR #916, opened 2026-04-02
feat: Add checkbashisms

emilal:
`checkbashisms` doesn't have the sort of project page people might expect, but there are at least some manpages available online, e.g. [checkbashisms(1) at die.net](https://linux.die.net/man/1/checkbashisms).

I gave [the Debian devscripts page](https://tracker.debian.org/pkg/devscripts) as the upstream because that's what it is, even though it's probably rather opaque for someone discovering it through the list of linters.

### DIFF
--- a/lua/lint/linters/checkbashisms.lua
+++ b/lua/lint/linters/checkbashisms.lua
@@ -1,16 +1,18 @@
 return {
   cmd = "checkbashisms",
-  stdin = false, -- Does not take stdin unless we give "-" as an arg
+  stdin = true, -- Only works if we give "-" as an arg
   append_fname = true,
   stream = "stdout",
   ignore_exitcode = true, -- exits > 0 if emits lints
   args = {
     "--lint",
+    "-",
   },
   -- checkbashism's manpage gives the following as the lint style output:
   -- {filename}:{lineno}:1: warning: possible bashism; {explanation}
   -- i.e. the column number and severity is entirely static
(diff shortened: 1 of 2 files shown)


lua/lint/linters/checkbashisms.lua
```diff
@@ -0,0 +1,17 @@
+return {
+  cmd = "checkbashisms",
+  stdin = false, -- Does not take stdin unless we give "-" as an arg
```
mfussenegger: Any reason not to set this to true and add the `-` arg?

emilal: Not really, I just don't know what your preference here is. I think either is fine. If you want I can swap it to stdin mode?

emilal: (If so I'm probably also going to replace `%f` in the parser with `(stdin)` since I see no need to spend effort parsing a static string)

emilal: I lined up a meta-PR so I can click the button here if that's the preferred way of doing things
https://github.com/emilal/nvim-lint/pull/1

mfussenegger: stdin is generally preferred as it is more flexible - allowing users to lint on modified buffers instead of only after writing the state to disk.

emilal: Okay, switched it to taking stdin

PR #917, opened 2026-04-05
feat!(checkpatch): implement stdin handling

dtor:
This change updates the checkpatch linter to support evaluating unsaved buffers so that it can run and provide updated results as soon as user exits insert mode.

Unfortunately kernel maintainer is against fixing checkpatch.pl's full file mode to properly handle stdin [1] so we have to work around this generating a diff against /dev/null and passing it to checkpatch instead.

The linter dynamically detects if the checkpatch script supports the '--ignore' flag and applies common ignores if available (some of them are too expensive for an online linter and are not changing often: "obsolete" and "SPDX check", while others do not make sense for our use case).

Unfortunately because now command is a "diff ... | checkpatch ..." it is hard to override it. The suggestion therefore is to make sure custom checkpatch (if needed) is first in $PATH.

Because linter code now returns function and not a table this change marked as breaking.

[1] https://lore.kernel.org/r/acTPXMJfkHLeItrK@google.com/

### DIFF
--- a/lua/lint/linters/checkpatch.lua
+++ b/lua/lint/linters/checkpatch.lua
@@ -1,48 +1,94 @@
 -- This favors the version of `checkpatch.pl` supplied by the project
 -- to which the file being linted belongs, falling back to a version
--- accessible via `$PATH`. If a user wishes to use a different version
--- of the script they can override "cmd" in their config:
--- require("lint").linters.checkpatch.cmd = '…/checkpatch.pl'
+-- accessible via `$PATH`.
 
-local tool_name = 'checkpatch.pl'
+local uv = vim.loop or vim.uv
 
--- path/to/file:line: severity: message
-local pattern = '([^:]+):(%d+): (%a+): (.+)'
-local groups = { 'file', 'lnum', 'severity', 'message' }
 local severity_map = {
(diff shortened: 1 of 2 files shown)


dtor:
Hi,

While I can replace vim.fs.joinpath() with a custom implementation maybe it would make sense to bump up minium configuration to 0.10.0 if not 0.11.0?

Also, what do you think about allowing extending lint.Linter class with linter-specific options? For example I would like to be able to adjust ignored checks list, depending on where in the kernel tree I am and with stdin handling it is not easy to reach "into" args building code. It would be straightforward to have them as an additional field.

Thanks!

mfussenegger:
> Unfortunately kernel maintainer is against fixing checkpatch.pl's full file mode to properly handle stdin [1]

What a bummer.

The changes here seem a bit too complex for my taste. I'd instead be inclined to add a `tmpfile` option or something like it to `try_lint`, so that if it's `true`, linters with `stdin = false` will generate a tmpfile during linting with the current buffer contents, and receive that as file argument. Probably limited to only do that if the buffer is actually modified. I guess a problem is that some linters likely require the file to also be in the cwd instead of `/tmp` - not sure yet how to deal with that.

That way the logic could be used for any `stdin = false` linter, without special per-linter logic.

PR #919, opened 2026-04-21
feat(linters): add dialyxir

ChrisAmelia:
* add [dialyxir](https://github.com/jeremyjh/dialyxir) for Elixir.

### DIFF
--- a/spec/dialyxir_spec.lua
+++ b/spec/dialyxir_spec.lua
@@ -1,0 +1,35 @@
+describe("linter.dialyxir", function()
+  it("can parse the output", function()
+    local parser = require("lint.linters.dialyxir").parser
+    local bufnr = vim.uri_to_bufnr("file:///foo.ex")
+    local result = parser( [[
+::warning file=foo.ex,line=16,col=11,title=unmatched_return::The expression produces multiple types, but none are matched.
+::warning file=foo.ex,line=28,title=invalid_contract::Invalid type specification for function should_fail.
+]], bufnr)
+    assert.are.same(2, #result)
+
+    -- -1 because rows and columns from 'mix dialyzer' are 1-based
+
+    local expected_error = {
+      col = 11 - 1,
+      lnum = 16 - 1,
+      severity = 2,
+      message = "The expression produces multiple types, but none are matched.",
+      source = "dialyzer",
+      bufnr = bufnr,
+    }
+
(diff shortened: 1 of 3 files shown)


spec/dialyxir_spec.lua
```diff
@@ -0,0 +1,35 @@
+describe("linter.dialyxir", function()
+  it("can parse the output", function()
+    local parser = require("lint.linters.dialyxir").parser
+    local bufnr = vim.uri_to_bufnr("file:///foo.ex")
+    local result = parser( [[
+::warning file=lib/server/lsp_server.ex,line=16,col=11,title=unmatched_return::The expression produces multiple types, but none are matched.
+::warning file=lib/lsp/diagnostic.ex,line=28,title=invalid_contract::Invalid type specification for function should_fail.
+]], bufnr)
+    assert.are.same(2, #result)
```
mfussenegger, line 9: Seems like the file filter in the parser is discarding the results.

ChrisAmelia: Replaced filename to match the declared file in `vim.uri_to_bufnr("file:///foo.ex")`.

```suggestion
    local bufnr = vim.uri_to_bufnr("file:///foo.ex")
    local result = parser( [[
::warning file=foo.ex,line=16,col=11,title=unmatched_return::The expression produces multiple types, but none are matched.
::warning file=foo.ex,line=28,title=invalid_contract::Invalid type specification for function should_fail.
]], bufnr)
    assert.are.same(2, #result)
```

ChrisAmelia:
I don't understand why the test fails with:

> spec/dialyxir_spec.lua:9: in function <spec/dialyxir_spec.lua:2>

I run the test locally with:

```sh
nvim --headless -c "PlenaryBustedDirectory spec/ {minimal_init}" > result.txt
```

spec/dialyxir_spec.lua
ChrisAmelia: > Seems like the file filter in the parser is discarding the results.

Removed the file filter, after testing with or without it, the result is the same: the correct diagnostics are displayed per buffer.

See https://github.com/mfussenegger/nvim-lint/compare/21dddcb0d822fdd93750ee2535beb594b6d49d2f..03fd379093a95c1804ced6ebd6595e2a6cf62942

PR #920, opened 2026-04-30
Fix golangci-lint reporting "undefined" errors in go.work environments

sunnuay:
When opening a session at the root of a Go workspace (go.work) where no go.mod exists, go env GOMOD returns /dev/null. This causes the script to fall back to single-file mode. As a result, golangci-lint only analyzes the current buffer, ignoring other files in the same package and triggering false positive "undefined" errors for cross-file variables.

Added a check for GOWORK. The script now verifies both GOMOD and GOWORK. It will only fall back to single-file mode if both are empty or /dev/null. This preserves the directory-level linting (:h) behavior required for Go workspaces to correctly resolve package-level variables.

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -21,11 +21,23 @@ local getArgs = function()
   if not ok then
     return
   end
+
+  local go_work_location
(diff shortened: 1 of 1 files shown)


sunnuay:
Related to #900

mfussenegger:
See also https://github.com/mfussenegger/nvim-lint/pull/939#issuecomment-4729999412

PR #928, opened 2026-05-20
clippy: file_name relative to lsp root dir (if available)

marjakm:
I tried to use rust clippy linter, but it did not show any clippy errors. It seems clippy json output `span.file_name` is always relative to the directory where `Cargo.toml` is located, but nvim-lint tries to compare it to `file_name` variable that is relative to cwd.

I don't know if there already is some way to fix the paths, but this is my attempt - just use the `root_dir` lsp has figured out, if available.

### DIFF
--- a/lua/lint/linters/clippy.lua
+++ b/lua/lint/linters/clippy.lua
@@ -38,7 +38,13 @@ return {
     local diagnostics = {}
     local items = #output > 0 and vim.split(output, "\n") or {}
     local file_name = vim.api.nvim_buf_get_name(bufnr)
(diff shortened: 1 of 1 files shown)


mfussenegger:
Root file heuristics are out of scope as they're to some degree subjective and it would significantly increase the maintenance burden. You can instead set the `cwd` on your `try_lint` call and infer it from LSP. See also https://github.com/mfussenegger/nvim-lint/issues/482#issuecomment-1999185606  

And there's also https://github.com/neovim/neovim/issues/34622

PR #931, opened 2026-05-28
oelint: Fix parsing of color enabled output

offa:
If color is enabled through oelint configuration the parser wont match and no issues are reported.

Stripping the ANSI color sequences from the linter output addresses this.

### DIFF
--- a/lua/lint/linters/oelint-adv.lua
+++ b/lua/lint/linters/oelint-adv.lua
@@ -7,6 +7,19 @@ local severity_map = {
   ['info'] = vim.diagnostic.severity.INFO,
 }
 
+local parser_core = require('lint.parser').from_pattern(
+  pattern, groups, severity_map,
+  { ['source'] = 'oelint-adv' }
+)
+
+local function strip_ansi_codes(s)
+  return s and s:gsub('\27%[[0-9;]*m', '') or s
+end
+
+local function parser(output, ...)
+  return parser_core(strip_ansi_codes(output), ...)
+end
+
 return {
   cmd = 'oelint-adv',
   stdin = false,
@@ -16,8 +29,5 @@ return {
   },
   ignore_exitcode = true,
   stream = 'stderr',
-  parser = require('lint.parser').from_pattern(
-    pattern, groups, severity_map,
-    { ['source'] = 'oelint-adv' }
-  ),
+  parser = parser,
 }
--- a/lua/lint/linters/oelint-adv.lua
+++ b/lua/lint/linters/oelint-adv.lua
@@ -14,6 +14,9 @@ return {
     '--quiet',
     '--messageformat={path}:{line}:{severity}:{id}:{msg}',
   },
+  env = {
+    ["NO_COLOR"] = "1",
+  },
   ignore_exitcode = true,
   stream = 'stderr',
   parser = require('lint.parser').from_pattern(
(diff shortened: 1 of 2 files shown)


spec/oelint-adv_spec.lua
```diff
@@ -2,13 +2,14 @@ describe('linter.oelint-adv', function()
   it('can parse the output', function()
     local parser = require('lint.linters.oelint-adv').parser
     local bufnr = vim.uri_to_bufnr('file:///foo.bb')
+    local escape_seq = string.char(27)
     local result = parser([[
 /foo.bb:1:error:oelint.var.mandatoryvar.HOMEPAGE:Variable 'HOMEPAGE' should be set [branch:true]
 /foo.bb:1:info:oelint.var.suggestedvar.CVE_PRODUCT:Variable 'CVE_PRODUCT' should be set [branch:true]
 /foo.bb:2:warning:oelint.vars.spacesassignment:Suggest spaces around variable assignment. E.g. 'FOO = "BAR"' [branch:true]
-]], bufnr)
+]] .. escape_seq .. "[31m/foo.bb:17:error:oelint.var.mandatoryvar.DESCRIPTION:Variable 'DESCRIPTION' should be set" .. escape_seq .. "[0m [branch:true]", bufnr)
```
offa: Getting the ANSI escape sequence right here is quite tricky. Just pasting it there didn't work as intended.

mfussenegger (commented):
Would it be possible to instead override the config with some kind of argument/flag, or setting a `NO_COLOR` env variable?

offa:
Unfortunately not. The CLI allows enable only and there are no env variables for config override.

mfussenegger:
Any chance to bring this up upstream to have an option to force no-colors? I'd like to avoid having to deal with this in nvim-lint. Other editor integrations likely have the same problem.

See also http://no-color.org/

offa:
`NO_COLOR` support was added in the new release :+1:. I have reverted my initial approach and set `NO_COLOR=1` instead.

While this works as expected when calling the parser manually, nvim-lint doesn't show any warnings. The `env` line sets `NO_COLOR` correctly, but it drops the other env vars. What’s the recommended approach here?

----

#### Update

Not touching anything at all, but just using master works now!? :thinking:

mfussenegger:
> but it drops the other env vars. 

That's currently sort of working as intended. It only preserves `PATH` if a linter has `env` set. Is that a problem?


> Not touching anything at all, but just using master works now!?

That's odd, can you investigate more?

offa:
> That's currently sort of working as intended. It only preserves PATH if a linter has env set. Is that a problem?

I need to look into that more carefully; that's at least how it seems.

> That's odd, can you investigate more?

Error on my end, configuration file wasn't picked up correctly.

lua/lint/linters/oelint-adv.lua
```diff
@@ -16,6 +16,7 @@ return {
   },
   env = {
     ["NO_COLOR"] = "1",
+    ["HOME"] = os.getenv("HOME"),
```
offa, line 19: The missing `HOME` caused the linter to crash because it relies on it. A naive fix is to restore it that way. Is there a better solution?

mfussenegger: I'm considering to copy the current env by default if any linter.env variables are set. But for now this seems like a good solution

PR #936, opened 2026-06-07
fix: suppress cd output when changing directories during linting

burkayanduv:
When nvim-lint changes directories to run linters with a different cwd, vim.cmd.cd() prints the directory path. This causes unwanted notifications on every lint operation.

Added silent = true to the cd command modifiers to suppress this output.


pg-fine:
Thank you for the fix, working great!

PR #938, opened 2026-06-10
Add csharpier linter

ilan-schemoul:
Adds [CSharpier](https://github.com/belav/csharpier) (the opinionated C# formatter, `cs` filetype) as a linter, reporting a WARN diagnostic when a file is not formatted.

## Notes on the implementation

- `csharpier check` only reports whether a file is formatted, plus a snippet of the expected formatting around the first difference. The parser emits a single diagnostic at the reported line with the expected snippet in the message (`Replace by:` + formatted code, visible in the diagnostic float).
- **stdin can't be used**: `csharpier check` with piped input silently behaves like `format` (prints formatted code, exits 0), so the file on disk is checked instead (`stdin = false`). Requires csharpier >= 1.0 (the 1.0 release changed the CLI from `dotnet-csharpier --check` to `csharpier check`).
- Findings go to **stderr** (stdout only carries a `Checked N files in Xms.` summary), and the exit code is 1 when unformatted, hence `stream = 'stderr'` and `ignore_exitcode = true`.
- `--no-msbuild-check` skips probing csproj files for a CSharpier.MsBuild version mismatch, which keeps the check fast and avoids spurious failures when linting a single buffer.

All output shapes were verified against csharpier 1.2.6. Spec included (`spec/csharpier_spec.lua`) covering the unformatted and clean cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,7 @@ Other dedicated linters that are built-in are:
 | [cppcheck][22]                         | `cppcheck`             |
 | [cpplint][cpplint]                     | `cpplint`              |
 | [credo][credo]                         | `credo`                |
+| [csharpier][csharpier]                 | `csharpier`            |
 | [dialyxir][dialyxir]                   | `dialyxir`             |
 | [cspell][36]                           | `cspell`               |
 | [cue][cue]                             | `cue`                  |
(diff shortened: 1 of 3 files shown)


mfussenegger:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Thanks but no

ilan-schemoul:
Oh you cannot review 40 lines of codes+tests. Interesting.

PR #939, opened 2026-06-11
Enable golangci-lint to run in subdirectories

arieroos:
I have this setup for a monrepo:

project/
-> app/ -- a fllutter project
-> backend/ -- a go project

As a result I can't tun golangcilint from nvim directly with this plugin.

This PR fixes that

### DIFF
--- a/lua/lint/linters/golangcilint.lua
+++ b/lua/lint/linters/golangcilint.lua
@@ -5,9 +5,37 @@ local severities = {
   convention = vim.diagnostic.severity.HINT,
 }
 
+local function with_cwd(cwd, fn, ...)
+  local curcwd = vim.fn.getcwd()
+
+  if curcwd == cwd then
+    return fn(...)
+  else
+    local mods = { noautocmd = true }
+    vim.cmd.cd({ cwd, mods = mods })
+    local ok, result = pcall(fn, ...)
+    vim.cmd.cd({ curcwd, mods = mods })
+    return ok, result
+  end
+end
+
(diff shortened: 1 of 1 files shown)


arieroos:
Fixes #900

mfussenegger:
Sorry but golangci-lint is already too complex as it is. Imho it would be worth getting some of these things upstream to have it detect different project configurations - to avoid having every editor integration deal with this

arieroos:
How unfortunate, the issue for this in golangci-lint's repo has been open for around 7 years

mfussenegger:
As a workaround you could probably run something like  `try_lint("golangcilint", { cwd = golang_cwd() }` with `golang_cwd` defined in your nvim config.

arieroos:
That's not a bad idea.

The golang_cwd() function could be along the lines of the getCwd() function in my fork.

Here is an untested version:
```lua
local golang_cwd = function()
   local curcwd = vim.fn.getcwd()
   local filename = vim.api.nvim_buf_get_name(0)

    local mods = { noautocmd = true }
    vim.cmd.cd({ cwd, mods = mods })
    local  ok, go_mod_location = pcall(dir, vim.fn.system, { 'go', 'env', 'GOMOD' })
    vim.cmd.cd({ curcwd, mods = mods })

  if not ok or go_mod_location == '/dev/null' or go_mod_location == '' then
    return curcwd
  end

  return vim.fn.fnamemodify(go_mod_location, ':h')
end
```

PR #947, opened 2026-07-27
Fix fortitude.lua parsing of line positions

vascocbf:
fortitude linter output uses 'line' not 'row'

changed: 
item.location.row --> item.location.line

Corrected the error for me


gomfol12:
Hi, good find. They changed it in this commit PlasmaFAIR/fortitude@0eda11975096517b16bf30bf683011f83a24e55e. Can you also update the tests in `spec/fortitude_spec.lua`?

vascocbf:
Thank you for the update @gomfol12. I think it should be fine now, just updated `spec/fortitude_spec.lua` to use `line` instead of `row`.

